PR #1, opened 2021-03-11
Convert to Symfony app

MusikAnimal:
Bug: T275546


MusikAnimal:
From here, the src/Engine directory can hold the services for other engines that we add. When that time comes, OcrController will require some reworking as it currently assumes we're only working with Google OCR.

CONTRIBUTING.md
```diff
@@ -0,0 +1,18 @@
...
+
+* [Node](https://nodejs.org) with the version specified by the `.nvmrc` [nvm](https://github.com/nvm-sh/nvm#installing-and-updating) file.
+* [Yarn](https://yarnpkg.com/getting-started/install) (`npm install -g yarn`)
+
+## Installation ##
+
+* `composer install`
+* Add the missing values from `.env` to a `.env.local` file
+* `symfony serve` to start the application
+* `yarn encore dev-server` if you need to make JS/CSS changes.
```
samwilson: Needs `npm install` before this, and actually I had to run encore before getting anything to work because otherwise it gave a "Could not find the entrypoints file from Webpack" exception.

I'm not sure how we want to handle this in production: for SVG Translate we put the built assets in the repo but it'd be nice to avoid that if possible. I guess we can install Node on the server and add a build step to the toolforge-bundle's deploy script?

Also, yarn isn't actually needed at this point, `npm run build` works fine (you've already added the scripts for build, server, etc.).

assets/app.js
```diff
@@ -0,0 +1,19 @@
+/*
+ * Welcome to your app's main JavaScript file!
+ *
+ * We recommend including the built version of this JavaScript file
```
samwilson: Remove this boilerplate stuff?

samwilson (approved):
This all looks totally good, and works in my testing.

My only real question is about how we plan to deploy the built assets, but that doesn't actually have to be resolved in this PR. I'd be happy to merge this as-is and iterate as required.

CONTRIBUTING.md
MusikAnimal: I used yarn to install bootstrap, but I suppose `npm install --save-dev bootstramp@3.3.6` would do the same thing? I'm wondering why Symfony recommends Yarn at all.

samwilson: I've wondered the same :)

This patch works fine for me without yarn. Happy to leave it in though, if you think it's easier to use yarn.

assets/app.js
MusikAnimal: Done

CONTRIBUTING.md
MusikAnimal: Yarn is no more :)

> I'm not sure how we want to handle this in production: for SVG Translate we put the built assets in the repo but it'd be nice to avoid that if possible. I guess we can install Node on the server and add a build step to the toolforge-bundle's deploy script?

I don't mind committing the compiled assets, just gotten used to it. It also ensures your local will match production. For whatever reason, my NVM doesn't always auto-change versions based on `.nvmrc`, so I sometimes build assets with the wrong version, which CI could pick up (like we did for SVG Translate). But that's my opinion :) We totally can get Node setup on our VPS. Not all apps have this build step so ToolforgeBundle may need to take that into account (fail gracefully, or maybe offer a `--encore` flag or something). Also most of us will use the Docker container which will ensure we test with Node 12.20.1, matching production. So yeah... what you propose sounds fine :)

samwilson: hmm I don't know now!

I've also gotten used to being told off by CI when I stuff up the built assets, and it's caught things a few times I think.

It feels a bit wrong, but does certainly make deployment easier. The Symfony-official way seems to be that one should build assets "somewhere" other than production and copy them over, but we don't have a 3rd place like that.

I'm happy to carry on checking built files into the repo. Pinging @dayllanmaza and @cimurah for your thoughts.

dayllanmaza: I don't know what's considered "best practices" here. With that said, compiling in production doesn't sound like the safest idea, and I would avoid installing anything not required by the application to run. If we were to do it anyway, I don't think ToolforgeBundle should handle it unless we add the post-deploy hooks to the bundle and add a separate script for building the files (not in ToolforgeBundle)

In any case, this project is small, and our deployment "steps" are very basic so I think it's fine to commit the compiled assets, at least for now.

cimurah: I agree that it feels wrong storing built assets in the repo since they are files that we don't really need track changes for, but it does make deployment easier or maybe we just got used to it.  I say we try not including assets in the repo and see that how that feels in our process.

PR #4, opened 2021-04-02
Add GitHub Actions

cimurah:
Add actions that will be run by GitHub

Bug: T277667

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,33 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    env:
+      APP_ENV: test
+    
+    strategy:
+      matrix:
+        php: [ '7.3', '7.4' ]
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    
+    - name: Set up PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{matrix.php}}
+        coverage: none
+
+    - name: Install wikimedia-ocr
+      run: |
+        composer install
+   
+    - name: Test
+      run: |
+        composer test
--- a/composer.json
+++ b/composer.json
@@ -62,6 +62,9 @@
         ],
         "post-update-cmd": [
             "@auto-scripts"
+        ],
+        "test": [
+            "./vendor/bin/phpcs -s -p ."
         ]
     },
     "conflict": {


.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    - name: Install wikimedia-ocr
+      run: |
+        composer install
+        npm run build
+
+    - name: Test
+      run: |
+        composer test
+        ./bin/phpunit --exclude-group=exclude-from-ci
```
samwilson: We don't yet have phpunit in use, so this will error. Also, `exclude-from-ci` is only something we do in WS Export because we haven't yet got all the tests passing; it's probably not something we'll aim to do here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+
+    env:
+      APP_ENV: test
+    
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
```
samwilson: If we're only testing on one OS, it doesn't need to be in the matrix. I don't know if we want to add other OSs; it doesn't seem very useful for this project, although making sure it'll run on Mac and Windows might be good for developers.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,45 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+      - dev
```
samwilson: We don't have a dev branch.

samwilson (changes requested):
Looks good. I'm not sure why the workflow isn't actually running. It looks like it did for one of the pushes to this branch, but nothing's showing up for this one (or am I looking in the wrong place). It should be possible to have this passing this before merging it to main.

.github/workflows/ci.yml
cimurah: We don't. Thank you!

.github/workflows/ci.yml
cimurah: Got it!

cimurah: Does this apply for php as well? Do we need different versions?

.github/workflows/ci.yml
cimurah: Done.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+
+    - name: Install wikimedia-ocr
+      run: |
+        composer install
+        npm run build
+
+    - name: Test
+      run: |
+        composer test
+        npm test
```
MusikAnimal: There is no `test` command defined in package.json, and I doubt we'll have one (unless we start writing JavaScript tests or something), so this can be removed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+        coverage: none
+
+    - name: Install wikimedia-ocr
+      run: |
+        composer install
+        npm run build
+
+    - name: Test
+      run: |
+        composer test
```
MusikAnimal, line 33: There's no `test` command defined in composer.json yet, but we should probably add one. So in this same PR you could add this to composer.json:
```
{
    ...
    "scripts": {
        ...
        "test": [
            "./vendor/bin/phpcs -s -p ."
        ]
    }
}
```

.github/workflows/ci.yml
MusikAnimal: Wikimedia OCR explicitly only works on PHP 7.3 right now. If we move to VPS we'll probably bump to 7.4, so I suppose it doesn't hurt to keep testing against both 7.3 and 7.4

.github/workflows/ci.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Set up PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{matrix.php}}
+        coverage: none
+
+    - name: Install wikimedia-ocr
+      run: |
+        composer install
+        npm run build
```
MusikAnimal: We do have a `build` script in our package.json, but it just compiles assets. Right now we only have PHP CodeSniffer tests which does not depend on assets being compiled, so CI doesn't need to be concerned with the Node dependency at all, I think. But even if it did, I assume you'd need to run `npm install` before running `npm run build`.

.github/workflows/ci.yml
cimurah: Sounds good. Thank you!

.github/workflows/ci.yml
cimurah: done. Thank you!

.github/workflows/ci.yml
cimurah: got it. Thanks!

.github/workflows/ci.yml
cimurah: I just deleted it for simplicity :)

PR #5, opened 2021-04-05
Add Tesseract as optional OCR engine

samwilson:
Bug: https://phabricator.wikimedia.org/T279118

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -5,7 +5,6 @@ namespace App\Controller;
 
 use App\Engine\EngineBase;
 use App\Engine\EngineFactory;
-use App\Engine\GoogleCloudVisionEngine;
 use App\Engine\TesseractEngine;
 use App\Exception\OcrException;
 use Krinkle\Intuition\Intuition;
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -3,7 +3,10 @@ declare(strict_types = 1);
 
 namespace App\Controller;
 
+use App\Engine\EngineBase;
+use App\Engine\EngineFactory;
 use App\Engine\GoogleCloudVisionEngine;
+use App\Engine\TesseractEngine;
 use App\Exception\OcrException;
 use Krinkle\Intuition\Intuition;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
@@ -17,7 +20,7 @@ class OcrController extends AbstractController
     /** @var Intuition */
     protected $intuition;
 
-    /** @var GoogleCloudVisionEngine */
+    /** @var EngineBase */
     protected $engine;
 
     /** @var mixed[] Output params for the view or API response. */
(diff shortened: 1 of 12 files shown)


src/Controller/OcrController.php
```diff
@@ -36,13 +38,19 @@ class OcrController extends AbstractController
...
+        EngineFactory $engineFactory
+    ) {
         // Dependencies.
         $this->intuition = $intuition;
-        $this->engine = new GoogleCloudVisionEngine($endpoint, $key);
+
+        $request = $requestStack->getCurrentRequest();
+
+        // Engine.
+        $this->engine = $engineFactory->get($request->get('engine'));
```
MusikAnimal: Should we set a default here, ala `$request->get('engine', 'google')`? Otherwise I get `Argument 1 passed to App\Engine\EngineFactory::get() must be of the type string, null given` when browsing to the home path

src/Controller/OcrController.php
```diff
@@ -36,13 +38,19 @@ class OcrController extends AbstractController
      * @param string $endpoint
      * @param string $key
```
MusikAnimal: Docblock needs updating following the change to the signature

MusikAnimal (changes requested):
Works in my testing! Just a few issues. Also could you update CONTRIBUTING.md with the Tesseract requirement? Interestingly without it installed, I get an error that I have the wrong version (version 1) – not that it's missing from my system. PR needs a rebase, too.

src/Engine/EngineBase.php
```diff
@@ -0,0 +1,30 @@
...
+namespace App\Engine;
+
+use App\Exception\OcrException;
+
+abstract class EngineBase
+{
+
+    /**
+     * @param string $imageUrl
+     * @param string $lang
```
MusikAnimal: `string|null` which I thought the linter would catch?

src/Controller/OcrController.php
samwilson: Done.

src/Controller/OcrController.php
samwilson: Oops! Yes, good idea. Done.

src/Engine/EngineBase.php
samwilson: Done.

I guess it's a MediaWiki codesniffer rule that catches that? But Phan finds it correctly now.

samwilson:
> Works in my testing! Just a few issues. Also could you update CONTRIBUTING.md with the Tesseract requirement? Interestingly without it installed, I get an error that I have the wrong version (version 1) – not that it's missing from my system. PR needs a rebase, too.

All done.

src/Engine/TesseractEngine.php
```diff
@@ -0,0 +1,48 @@
...
+            throw new OcrException('image-retrieval-failed', [$exception->getMessage()]);
+        }
+
+        // Sanitize the language code.
+        $cleanLang = preg_replace('/[a-zA-Z]+/', '', $lang);
+
+        // Run OCR.
+        $ocr = new TesseractOCR();
+        // @TODO Remove this after this is resolved: https://github.com/thiagoalessio/tesseract-ocr-for-php/issues/210
+        FriendlyErrors::checkTesseractPresence($ocr->command->executable);
```
samwilson, line 40: This means it runs `type tesseract` twice before actually running it, which seems a bit unnecessary to do on every single run. I might add a patch upstream.

PR #6, opened 2021-04-05
Add basic docker env

dayllanmaza:
Bug: https://phabricator.wikimedia.org/T277668

### DIFF
--- a/docker/run.sh
+++ b/docker/run.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-docker run -p ${DOCKER_OCR_PORT:-8000}:8000 --mount type=bind,source="$(cd "$(dirname $0)/.."; pwd)",target=/wikimedia-ocr wikimedia-ocr:latest
+docker run -ti -p ${DOCKER_OCR_PORT:-8000}:8000 --mount type=bind,source="$(cd "$(dirname $0)/.."; pwd)",target=/wikimedia-ocr wikimedia-ocr:latest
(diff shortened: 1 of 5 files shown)


docker/run.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+docker run -p ${DOCKER_OCR_PORT:-8000}:8000 --mount type=bind,source="$(pwd)",target=/wikimedia-ocr wikimedia-ocr:latest
```
samwilson: This could have `source="$(cd "$(dirname $0)/.."; pwd)"` to guard against when it's run from a different directory.

samwilson (commented):
I haven't fully got this working yet, but just noticed this one thing.

docker/run.sh
dayllanmaza: Thanks

MusikAnimal (commented):
Works in my testing! However I'm not sure how to _stop_ the container. `./docker/run.sh` spins up the symfony server and I see the output, but the usual Ctrl+C doesn't kill it. I actually had to run `kill -9` on the process.

dayllanmaza:
> Works in my testing! However I'm not sure how to _stop_ the container. `./docker/run.sh` spins up the symfony server and I see the output, but the usual Ctrl+C doesn't kill it. I actually had to run `kill -9` on the process.

It should handle Ctrl+C now. Alternatively, you can also run `docker ps` and `docker stop|kill <id or name>` to stop the container

PR #7, opened 2021-04-07
Add phan, linting, and minus-x to composer test

samwilson:


samwilson:
The uses of `@phan-suppress-next-line PhanTypeMismatchDimFetch` here will most likely be fixed by T277129.

PR #10, opened 2021-04-07
Add tesseract support to docker

cimurah:
Add Tesseract ocr engine to docker.

Bug: T279553

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,7 @@ RUN apt-get update -q && apt-get install -y \
 	libicu-dev \
 	libzip-dev \
 	unzip \
+      tesseract-ocr \
       && docker-php-ext-install intl \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -11,6 +11,8 @@ RUN apt-get update -q && apt-get install -y \
 	libicu-dev \
 	libzip-dev \
 	unzip \
+      tesseract-ocr \
+      libtesseract-dev \
```
dayllanmaza: Do we need `libtesseract-dev`?

cimurah: It wasn't working for whatever reason without this lib. I removed and it worked fine. Thanks!

samwilson: Could possibly use `tesseract-ocr-all` instead, or maybe add some select language packages? I think by default it's just osd and eng, but we'll want to test on other languages.

PR #12, opened 2021-04-08
Improve validation of image URLs and error handling

MusikAnimal:
This further validates image URLs, and captures 404s from GoogleCloudVision

Bug: T279610

### DIFF
--- a/.phpcs.xml
+++ b/.phpcs.xml
@@ -5,6 +5,7 @@
         <properties>
             <property name="rootNamespaces" type="array">
                 <element key="src" value="App" />
+                <element key="tests" value="App\Tests" />
             </property>
         </properties>
     </rule>
@@ -14,5 +15,6 @@
     <exclude-pattern>./var/</exclude-pattern>
     <exclude-pattern>src/Kernel.php</exclude-pattern>
     <exclude-pattern>./node_modules/</exclude-pattern>
+    <exclude-pattern>./bin/.phpunit/</exclude-pattern>
     <exclude-pattern>./public/build/</exclude-pattern>
 </ruleset>
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -7,6 +7,7 @@ use App\Exception\OcrException;
 
 abstract class EngineBase
 {
+    public const ALLOWED_FORMATS = ['png', 'jpeg', 'jpg', 'gif', 'tiff', 'tif', 'webp'];
 
     /**
      * @param string $imageUrl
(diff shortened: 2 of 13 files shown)


samwilson:
I think you forgot to add the `./bin/phpunit` file. :-)

MusikAnimal:
> I think you forgot to add the `./bin/phpunit` file. :-)

Ha! And the tests, for that matter! That would explain the build failure :stuck_out_tongue: 

Fixed and I think this is ready for review now.

.phpcs.xml
```diff
@@ -5,14 +5,16 @@
         <properties>
             <property name="rootNamespaces" type="array">
                 <element key="src" value="App" />
+                <element key="tests" value="App\Tests" />
             </property>
         </properties>
     </rule>
 
-    <file>.</file>
+    <file>src/</file>
```
samwilson: This now excludes the new `tests/` directory, as well as `config/` etc.

I think leaving it as dot is fine, as you're excluding the `.phpunit/` directory below.

src/Engine/EngineBase.php
```diff
@@ -7,6 +7,7 @@
 
 abstract class EngineBase
 {
+    public const ALLOWED_FORMATS = ['svg', 'png', 'jpeg', 'jpg', 'gif', 'tiff', 'tif', 'webp'];
```
samwilson: Neither Google nor Tesseract support SVG, it looks like. We send PNG renderings of Commons SVGs instead.

src/Engine/GoogleCloudVisionEngine.php
```diff
@@ -43,6 +44,14 @@ public function getText(string $imageUrl, ?string $lang = null): string
             $this->gcv->setImage($imageUrl);
         } catch (LimitExceededException $e) {
             throw new OcrException('limit-exceeded', [$e->getMessage()]);
+        } catch (ErrorException $e) {
+            if (false !== strpos($e->getMessage(), '404 Not Found')) {
+                // The 'image-retrieval-failed' message is the important part that is localized.
+                throw new OcrException('image-retrieval-failed', ['404 Not Found']);
```
samwilson, line 50: I think we might want to add something to the API response that unifies the error messages of Google vs Tesseract, but that can come later. As it is, they give slightly different responses.

samwilson (changes requested):
Looks great! Works in all my testing.

src/Engine/GoogleCloudVisionEngine.php
MusikAnimal: Yeah, this was a bit of a hack trying to detect a simple 404 error. The GoogleCloudVision obscures any HTTP issue by running `file_get_contents` so the best we get is an `ErrorException` and plain text with the response code. I thought about maybe doing a HEAD request (since we don't need to fetch the image for Google) to the image in `EngineBase::checkImageUrl()`, that way we can get more meaningful response codes and return them in a consistent manner. How does that sound?

samwilson: I think with T277129 we'll be able to provide a URL directly, and not download the image file. In that case, when the URL doesn't exist we'll get back something like:

```
  -error: array:2 [▼
    "code" => 3
    "message" => "The URL does not appear to be accessible by us. Please double check or download the content and pass it in."
  ]
```

So maybe we just don't worry about it for now! :)

MusikAnimal: Okay! Glad I asked :)

.phpcs.xml
MusikAnimal: Done

src/Engine/EngineBase.php
MusikAnimal: Removed

PR #14, opened 2021-04-13
Switch Google Cloud Vision package

samwilson:
Our Wikisource fork of the old package is no longer maintained,
and Google maintains their own now.

Bug: T277129


CONTRIBUTING.md
```diff
@@ -13,7 +13,10 @@ If you need to make asset changes:
 * `composer install`
 * `npm install`
 * Add the missing values from `.env` to a `.env.local` file
-  * Use `https://vision.googleapis.com/` as the `APP_GOOGLE_CLOUD_VISION_ENDPOINT`, with your own [Cloud Vision API](https://cloud.google.com/vision) key as the `APP_GOOGLE_CLOUD_VISION_KEY`. Google gives you 1,000 free lookups per month.
+  * Create a new Google service account at https://console.cloud.google.com/iam-admin/serviceaccounts Google gives you 1,000 free lookups per month.
+  * Give the service account the *Compute Engine Service Account* role.
+  * Add a new key for the service account (of type JSON) and save it somewhere.
```
MusikAnimal: I'm assuming tests/fixtures/google-account-keyfile.json is an example of this. I wasn't sure where to get the `private_key_id`, and whether the `client_email` and `client_id` are required. I tried to guess my way through it and am getting the error `Warning: openssl_sign(): supplied key param cannot be coerced into a private key` when attempting to transcribe with Google Cloud Vision. Is there documentation on the structure of this keyfile somewhere?

When I started from scratch with a new service account, it seems the 'Compute Engine Service Account' role (and most all of the roles) are already "enabled". Is that expected?

samwilson: Sorry, good point. The entire JSON file is actually provided in the Google UI, and it doesn't need to be edited. I'll update the instructions to make this clearer.

samwilson: > When I started from scratch with a new service account, it seems the 'Compute Engine Service Account' role (and most all of the roles) are already "enabled". Is that expected?

I'm not really sure. I had to select them, so maybe there are different defaults for different projects or something? I must admit, I'm not totally across how all of the IAM stuff works.

CONTRIBUTING.md
```diff
@@ -13,7 +13,10 @@ If you need to make asset changes:
 * `composer install`
 * `npm install`
 * Add the missing values from `.env` to a `.env.local` file
-  * Use `https://vision.googleapis.com/` as the `APP_GOOGLE_CLOUD_VISION_ENDPOINT`, with your own [Cloud Vision API](https://cloud.google.com/vision) key as the `APP_GOOGLE_CLOUD_VISION_KEY`. Google gives you 1,000 free lookups per month.
+  * Create a new Google service account at https://console.cloud.google.com/iam-admin/serviceaccounts Google gives you 1,000 free lookups per month.
+  * Give the service account the *Compute Engine Service Account* role.
```
MusikAnimal, line 18: I think there's one more step here, which is to enable the Cloud Vision API. The error I got back from Google gave me a direct link to enable it, but https://console.cloud.google.com/apis/api/vision.googleapis.com/overview should also point to the right place (assuming they're logged in as the right service account)

MusikAnimal (approved):
Sorry for taking so long to review this! I did finally get it working. Instructions look good except I think we're missing the step to enable the Cloud Vision API. As I said Google will give you a direct link to enable it in the error response, so I'm fine with merging this as-is if you want, since it's pretty easy to figure it out.

cimurah (changes requested):
We need to add `&& docker-php-ext-install bcmath \` to `Dockerfile`.

CONTRIBUTING.md
samwilson: Good point. Done.

samwilson (commented):
> We need to add && docker-php-ext-install bcmath \ to Dockerfile.

Done.

MusikAnimal (approved):
Going to merge. I got in working in Docker, but I had to use a relative path for APP_GOOGLE_KEYFILE. For instance putting it in the vendor/ directory (only because that's git-ignored), the path would be `../vendor/name-of-keyfile.json`.

PR #16, opened 2021-04-16
Make it possible to provide multiple language codes

samwilson:
This adds a new ?langs[]=x parameter with which to provide
multiple language codes. It does some basic sanitization of the
input but doesn't compare with what any engine supports.

The existing ?lang=yy form is kept for backwards compatibility.

Bug: https://phabricator.wikimedia.org/T280214


src/Engine/TesseractEngine.php
```diff
@@ -30,14 +35,13 @@ public function getText(string $imageUrl, ?string $lang = null): string
...
 
-        // Sanitize the language code.
-        $cleanLang = preg_replace('/[a-zA-Z]+/', '', $lang);
-
         // Run OCR.
         $ocr = new TesseractOCR();
         $ocr->imageData($imageContent, $imageResponse->getHeaders()['content-length'][0]);
-        if ($cleanLang) {
-            $ocr->lang($cleanLang);
+        if ($langs && count($langs) > 0) {
```
MusikAnimal, line 41: `$langs` can only be null or an array, so I think just `if ($langs)` is sufficient, since an empty array is falsy. Minor detail, though, and one could argue `count($langs) > 0` is more clear so feel free to ignore me, hehe

src/Engine/TesseractEngine.php
```diff
@@ -30,14 +35,13 @@ public function getText(string $imageUrl, ?string $lang = null): string
...
+            // The Option::lang() method takes language codes as method arguments (not an array).
+            // @phan-suppress-next-line PhanUndeclaredMethodInCallable
+            call_user_func_array([$ocr, 'lang'], $langs);
```
MusikAnimal: Will `$ocr->lang(...$langs)` work? Just a tiny bit cleaner, in my opinion

templates/output.html.twig
```diff
@@ -7,11 +7,19 @@
...
         </div>
-        <div class="form-group">
-            <label for="lang" class="form-label">{{ msg('language-code') }}</label>
-            <input name="lang" id="lang" class="form-control" value="{{ app.request.get('lang') }}" />
-            <p class="help-block">{{ msg( 'language-code-help' ) }}</p>
-        </div>
+        <ol>
+            {# Add an additional blank language, to create an extra form field. #}
+            {% for lang_num,lang in langs|merge(['']) %}
+                {# @TODO Use a better system of selecting and displaying multiple languages. #}
```
MusikAnimal, line 13: It looks kinda weird when you browse to the form without any langs in the URL params, but as you say there's a TODO to improve this. Once we have a way to get a list of the supported languages, I might suggest using https://select2.org/getting-started/basic-usage#multi-select-boxes-pillbox

MusikAnimal (approved):
Two highly opinionated nitpicky things, otherwise looks great! Happy to merge this as-is

README.md
```diff
@@ -31,14 +31,23 @@ And get back a JSON response with either 'text' or 'error' top-level items set:
...
 Note that you should only set the `lang` parameter for languages that require it.
 The [documentation](https://cloud.google.com/vision/reference/rest/v1/images/annotate#imagecontext) informs us of the following:
 
 > In most cases, an empty value yields the best results since it enables automatic language detection.
 > For languages based on the Latin alphabet, setting languageHints is not needed.
 > In rare cases, when the language of the text in the image is known, setting a hint will help get better results
 > (although it will be a significant hindrance if the hint is wrong).
-> Text detection returns an error if one or more of the specified languages is not one of the supported languages.
+> Text detection returns an error if one or more of the specified languages is not
+> one of the [supported languages](https://cloud.google.com/vision/docs/languages).
```
cimurah, line 46: My tiny comment is that I didn't get any error in the UI when I specified an invalid language code.  Not sure if this patch is supposed to throw an error, but it doesn't matter much since we'll be adding language validation very soon.

cimurah (approved):
Works in my testing.

src/Engine/TesseractEngine.php
samwilson: That's what I thought too, but recently I've had people comment that it's better to be explicit like this. I don't really mine either way, but it is maybe easier to read with the `count()`.

README.md
samwilson: Oh interesting. You don't get any error when using `en` with Tesseract? But yeah, we can sort out all this when we add the validation.

src/Engine/TesseractEngine.php
samwilson: Yes! That's much clearer, thanks.

README.md
MusikAnimal: Using `en` with Tesseract I get an ErrorException: `Notice: fwrite(): write of 8192 bytes failed with errno=32 Broken pipe`

cimurah: Sorry for the confusion. I was trying to trigger an error using  Google since the reference is to the Google languages. 

For `Tesseract`, I'm also getting the Symfony Exception page with the message `Notice: fwrite(): write of 8192 bytes failed with errno=32 Broken pipe` when I submit an invalid language.

PR #21, opened 2021-04-28
Improve language selection by using Select2 library

MusikAnimal:
For now, this only allows you to manually input each language. In the
future, once we have an API of sorts to give us all valid languages for
the chosen OCR engine, we can use that to restrict Select2 to only
allowing those languages, and also having autocomplete.

This PR also commits package-lock.json and removes yarn.lock which we
don't recommend using anymore. Also the dev-server doesn't appear to
work correctly, but it's not that important anyway, so we'll just
instruct devs to use 'npm run watch'.

Also removed the 'form-label' class from <label> elements, which doesn't
do anything.

Bug: T280214


assets/app.js
```diff
@@ -1,13 +1,25 @@
...
     });
+
+    // For the result page. Makes the 'Copy' button copy the transcription to the clipboard.
+    const copyButton = $('#copy-button');
+    if (copyButton.length) {
+        copyButton.on('click', () => {
+            const textarea = $('#text');
+            textarea.select();
+            document.execCommand('copy');
+            copyButton.innerText = copyButton.dataset.copiedText;
```
samwilson: `copyButton` is now a jQuery object so should be `$copyButton` and use `$copyButton.text($copyButton.data('copied-text'));`. As it is, we get "Uncaught TypeError: e.dataset is undefined".

samwilson (changes requested):
This looks great. Thanks for improving my clunky attempt at this!

Just a little jQuery bug, but other than that this looks ready to go.

assets/app.js
MusikAnimal: Done! Thanks for noticing that!

samwilson (approved):
Looks good.

PR #22, opened 2021-04-28
Add support for PSM and OEM Tesseract options

MusikAnimal:
This also includes validations for the new options. In order to use
OcrException outside EngineBase::getText(), we replace the try/catch
approach with an exception listener. Consequently default parameter
values known only to the controller have to be exposed as a public
static property.

Bug: T280213


MusikAnimal:
This unnecessarily includes the frontend bits, too (I didn't read the ticket properly!), but still has some of the backend needed for the API.

I thought about having `TesseractEngine` have setters like `setPsm()` but we both instantiate the OCR engine and do the OCRing all in `EngineBase::getText()` right now. So I thought to add an `array $extras` to the signature, to leave room for whatever options other engines might have? I have other ideas, too.

src/Engine/GoogleCloudVisionEngine.php
```diff
@@ -28,10 +28,11 @@ public function __construct(string $keyFile)
      * Get transcribed text from the given image.
      * @param string $imageUrl
      * @param string[]|null $langs
+     * @param array|null $_extras Not used by the GoogleCloudVision engine.
      * @return string
      * @throws OcrException
      */
-    public function getText(string $imageUrl, ?array $langs = null): string
+    public function getText(string $imageUrl, ?array $langs = null, ?array $_extras = null): string
```
samwilson: Typo underscore in `$_extras`.

templates/_tesseract_options.html.twig
```diff
@@ -0,0 +1,23 @@
+<fieldset id="tesseract-fields">
+    <legend>{{ msg('tesseract-options') }}</legend>
+    <div class="form-group">
+        <label for="psm">{{ msg('tesseract-psm-label') }}</label>
+        <select name="psm" id="psm" class="form-control">
+            {# Construct the array so that the default (3) comes first. #}
+            {% for i in 0..13 %}
+                {# Each i18n message is 'tesseract-psm-N' where N is the PSM value. #}
```
samwilson: Should we do the MediaWiki thing of listing out all messages in full, for easier grepping?

samwilson (commented):
> I thought about having TesseractEngine have setters like setPsm() but we both instantiate the OCR engine and do the OCRing all in EngineBase::getText() right now. So I thought to add an array $extras to the signature, to leave room for whatever options other engines might have? I have other ideas, too.

I like the idea of setters like `TesseractEngine::setPsm()`, and calling them in the controller. Maybe all optional options (not to be tautological or anything) should be separate methods — i.e. add `setLangs()`. This would mean the engine objects would have a bunch of setters, but at any point it'd still be possible to call `getText($image)` and it still make sense.

We've already got a conditional in `OcrController::__construct()` for google vs tesseract, so maybe that can be expanded (from a ternary) and contain all the engine-specific bits?

src/Engine/GoogleCloudVisionEngine.php
MusikAnimal: That was to indicate `$extras` isn't used in this method, we just had to have it as an arg due to the signature in the abstract EngineBase. But, I'll go with your suggestion of having setters instead of an `$extras` arg.

templates/_tesseract_options.html.twig
MusikAnimal: Yes! Good idea.

MusikAnimal:
@samwilson Let me know what you think now. As suggested I went the route of adding Engine-specific setters, set in the controller. However for validation purposes, I want to throw an `OcrException` when currently we're only doing our try/catch when calling `getText()`. So instead, I created an `ExceptionListener` so we can throw `OcrException`s anywhere and they should be properly handled. The drawback is I lose all contextual bits from the controller (such as `$this->params`). To get around this, I changed `OcrController` to have a [static class property of defaults](https://github.com/wikimedia/wikimedia-ocr/pull/22/files#diff-02887fb256410c2b4193f1f7759ddcbbf2e6b33594416c5adeae559b988816edR26-R37). The URL query params get merged into them inside the `ExceptionListener`. The static class property is what I'm not completely happy with, but I'm not sure about alternative ways to do this since we can't magically 'resume' execution in the controller after intercepting an exception in an event listener. On the bright side, our exception handling logic lives in one place, and also sort of like how we can set default values for params in one place `OcrController::$params`).

src/Controller/OcrController.php
```diff
@@ -20,11 +20,21 @@ class OcrController extends AbstractController
     /** @var Intuition */
     protected $intuition;
 
-    /** @var EngineBase */
+    /** @var TesseractEngine|GoogleCloudVisionEngine */
```
MusikAnimal, line 22: Had to do this to appease Phan, otherwise it complains about calling the Engine-specific setters like `setPsm()`

src/EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,75 @@
...
+    public function onKernelException(ExceptionEvent $event): void
+    {
+        $exception = $event->getThrowable();
+
+        // We only care about OcrExceptions.
+        if (!$exception instanceof OcrException || !$event->isMasterRequest()) {
+            return;
+        }
+
+        $isApi = str_contains($this->request->getPathInfo(), 'api.php');
```
MusikAnimal, line 53: Fun discovery: Symfony comes with PHP 8 polyfills!

samwilson: Oh, TIL! :-)

src/Controller/OcrController.php
```diff
@@ -20,11 +19,21 @@ class OcrController extends AbstractController
...
 
-    /** @var mixed[] Output params for the view or API response. */
-    protected $params = [];
+    /**
+     * The output params for the view or API response.
+     * This also serves as where you define the defaults.
+     * Note this is used by the ExceptionListener.
+     * @var mixed[]
+     */
+    public static $params = [
```
samwilson, line 31: Should we look at adding the `dpi` option? We currently get warnings on most images:

> Warning: Invalid resolution 0 dpi. Using 70 instead.

I'm happy to hold that over for a future patch, but will it work with this system? It won't have a max value (well, I guess it will in practice, but it's a different sort of value).

samwilson (approved):
Looks great to me, works well. I'm happy to merge as-is, but just wondering about the DPI option.

src/Controller/OcrController.php
MusikAnimal: From playing around, it seems there is a valid DPI range from 70 to 2400. I added a DPI option, but depending on what other options you use, you still get the same error. For instance, with a PSM of 0, and OEM of 3, and DPI of 300, I see:
```
Error! The command did not produce any output.

Generated command:
"tesseract" - "/tmp/ocrXMeZIy" --psm 0 --oem 3 --dpi 300

Returned message:
Tesseract Open Source OCR Engine v4.1.1 with Leptonica
Warning. Invalid resolution 0 dpi. Using 70 instead.
```

Seems like there might be an issue with thiagoalessio/tesseract_ocr. But either way, I have no idea what options require you to set a DPI, and I suspect many of our users won't, either. Do we know what the conditions are where DPI is needed?

samwilson: Hmm yeah good points. I think the DPI is probably most useful for things like newspapers, which might also have a scan resolution set in the Index page. Let's leave it for now.

PR #23, opened 2021-04-28
Validate language codes

cimurah:


i18n/en.json
```diff
@@ -17,5 +17,6 @@
     "image-retrieval-failed": "Image retrieval failed: $1",
     "documentation": "Documentation",
     "version": "Version $1",
-    "report-issue": "Report an issue"
+    "report-issue": "Report an issue",
+    "langs-param-error": "This language is NOT supported by the OCR."
```
MusikAnimal: We'll want to indicate _which_ languages are invalid (there could be multiple), so something like `The following languages are not supported by the OCR engine: $1`. $1 would be a comma-separated list of the invalid languages. You can use Intuition's [`listToText()`](https://github.com/Krinkle/intuition/blob/39cc5fbb3ced1d1bff166bbe3434b5f1953bdd06/src/Intuition.php#L1416) to make that localized (because not every language uses commas). I think you'll have to inject the Intuition service to get access to that method.

src/Engine/TesseractEngine.php
```diff
@@ -28,25 +32,43 @@ public function getText(string $imageUrl, ?array $langs = null): string
...
+    {
+        $validLanguages = $this->ocr->availableLanguages();
+
+        if ($langs) {
+            foreach ($langs as $lang) {
+                if (!in_array($lang, $validLanguages)) {
+                    throw new OcrException('langs-param-error', []);
+                }
+            }
+        }
```
MusikAnimal: Should be able to simplify this to something like (let's assume each engine has a `getValidLangs()` method, see my other comments):
```
$invalidLangs = array_diff($langs, $this->getValidLangs());
if (count($invalidLangs)) {
    throw new OcrException('langs-param-error', $invalidLangs);
}
```

MusikAnimal (commented):
I know this is just a draft but assuming it's an invitation for early feedback... :)

Making an abstract method in `EngineBase` such that each engine must define it sounds like the right approach. However, the logic of _how_ to validate the language input is the same regardless of the engine. The only difference is _which_ languages are supported, right? So instead of an abstract `validateLangs()`, we could have an abstract `getValidLangs()` which returns the valid langs for that engine. Then EngineBase could have a normal public method `validateLangs(array $givenLangs)` that we call below `$this->checkImageUrl()` in `getText()` for each engine, if that makes sense. Just an idea!

There's of course the issue of mapping the input since we're given ISO 639-1 but Tesseract uses a different standard. I assume that's going to happen in this PR too? `TesseractEngine::getValidLangs()` could do that transformation, so that it returns ISO 639-1.

If you're not handling the mapping in this PR, then I think your approach of an abstract `validateLangs()` is fine :) The validation logic itself isn't very complicated, anyway.

src/Engine/EngineBase.php
```diff
@@ -29,4 +35,18 @@ public function checkImageUrl(string $imageUrl): void
...
         }
     }
+
+    /**
+     * @param string[]|null $langs
+     * @throws OcrException
+     */
+    public function validateLangs(?array $langs): void
+    {
+        $intuition = new Intuition();
```
samwilson: The Intuition service should be injected instead (which is what I've done in #24, so if that lands first you can just use `$this->getIntuition()`).

i18n/en.json
cimurah: Thank you for the pointers!! Done.

src/Engine/TesseractEngine.php
cimurah: Good point. Thank you!

src/Engine/EngineBase.php
cimurah: Good point. Thank you!

cimurah: I injected `Intuition` in the `EngineBase` class constructor, we can review and consolidate the best approach.

i18n/en.json
```diff
@@ -17,5 +17,6 @@
     "image-retrieval-failed": "Image retrieval failed: $1",
     "documentation": "Documentation",
     "version": "Version $1",
-    "report-issue": "Report an issue"
+    "report-issue": "Report an issue",
+    "langs-param-error": "The following language(s) are not supported by the OCR engine: $1"
```
samwilson: This could use `PARSE`, e.g.:

```
The following {{PLURAL:$1|language is|languages are}} not supported by the OCR engine: $2
```
with the invocation changed to something like:

```php
        if (count($invalidLangs)) {
            $langs = array_values($invalidLangs);
            throw new OcrException('langs-param-error', [ count($langs), $this->intuition->listToText($langs) ]);
        }
```

samwilson (approved):
Works well in my testing. Just a comment about the message, but that could also be done as a followup.

i18n/en.json
cimurah: Done. Thank you!

PR #24, opened 2021-04-29
Allow other image hosts, and cross origin API requests

samwilson:
* Set CORS header.
* Add new .env variable for setting allowed domain names of images.
* Remove all mentions of upload.wikimedia.org

Bug: https://phabricator.wikimedia.org/T280953


src/Engine/EngineBase.php
```diff
@@ -7,26 +7,45 @@
...
     public function checkImageUrl(string $imageUrl): void
     {
+        $hostRegex = implode('|', array_map('preg_quote', $this->getImageHosts()));
         $formatRegex = implode('|', self::ALLOWED_FORMATS);
-        $matches = preg_match("/^https?:\/\/upload\.wikimedia\.org\/.*($formatRegex)$/", strtolower($imageUrl));
+        $regex = "/https?:\/\/($hostRegex).*($formatRegex)$/";
+        $matches = preg_match($regex, strtolower($imageUrl));
         if (1 !== $matches) {
-            throw new OcrException('image-url-error', ['https://upload.wikimedia.org']);
+            throw new OcrException('image-url-error', [join(', ', $this->getImageHosts())]);
```
samwilson: This should also use `Intuition::listToText()` but I'm not quite sure what the best way to do it is.

It's probably easiest to just give EngineBase a Intuition member variable, but that feels slightly hacky.

samwilson:
Allowing all requests to the API is okay I think because we're still limiting it to only Wikimedia images. It's what's suggested at https://wikitech.wikimedia.org/wiki/News/Toolforge.org#Cross-Origin_Resource_Sharing_%28CORS%29_requests_broken

Even if we limit it to only Wikisource domains, any other script can still hit the API without trouble.

src/Engine/EngineBase.php
cimurah: I injected the Intuition member variable in https://github.com/wikimedia/wikimedia-ocr/pull/23/files so if that gets approved then you can just use the `Intuition::listToText()` :)

PR #28, opened 2021-05-04
Setup error emails

dayllanmaza:
Bug: https://phabricator.wikimedia.org/T279609


samwilson (approved):
Looks good, works in my tests.

I'm not sure why `package.json` is changed, but it seems to be a sensible clean up so no worries.

PR #29, opened 2021-05-06
Add API for available langs and autocompletion to lang selector

MusikAnimal:
New API endpoint is at /api/available_langs. For consistency /api can
now be accepted instead of /api.php

Also renamed controller methods to have 'Action' in the name to
distinguish which ones have routes.

Refactored constructing the API JsonResponse object.

Bug: https://phabricator.wikimedia.org/T282073

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -7,11 +7,11 @@ const $ = require('jquery');
 $(function () {
     // Initiate Select2, which allows dynamic entry of languages.
     $('#lang').select2({
-        tags: true,
         theme: 'bootstrap',
     });
 
     // Show engine-specific options.
+    // TODO: Update the Select2 options with available languages.
     $('#engine').on('change',  e => {
         $('.engine-options').addClass('hidden');
         $(`#${e.target.value}-options`).removeClass('hidden');
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -58,7 +58,7 @@ class OcrController extends AbstractController
         // Parameters.
         $this->imageUrl = (string)$request->query->get('image');
         static::$params['langs'] = $this->getLangs($request);
-        static::$params['image_hosts'] = $this->intuition->listToText($this->engine->getImageHosts());
+        static::$params['image_hosts'] = $this->engine->getImageHosts();
 
         $this->setEngineOptions($request);
     }
@@ -108,8 +108,15 @@ class OcrController extends AbstractController
      * @Route("/", name="home")
      * @return Response
      */
-    public function home(): Response
+    public function homeAction(): Response
     {
+        // Pre-supply available langs for autocompletion in the form.
+        static::$params['available_langs'] = $this->engine->getValidLangs();
+        sort(static::$params['available_langs']);
+
+        // Intution::listToText() isn't available via Twig, and we only want to do this for the view and not the API.
+        static::$params['image_hosts'] = $this->intuition->listToText(static::$params['image_hosts']);
+
         if ($this->imageUrl) {
             static::$params['text'] = $this->engine->getText($this->imageUrl, static::$params['langs']);
         }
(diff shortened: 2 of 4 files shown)


src/Controller/OcrController.php
```diff
@@ -58,7 +58,7 @@ public function __construct(RequestStack $requestStack, Intuition $intuition, En
         // Parameters.
         $this->imageUrl = (string)$request->query->get('image');
         static::$params['langs'] = $this->getLangs($request);
-        static::$params['image_hosts'] = $this->intuition->listToText($this->engine->getImageHosts());
+        static::$params['image_hosts'] = $this->engine->getImageHosts();
```
MusikAnimal, line 61: I changed this so that the API endpoint returns an array instead of a string. In the error message it still uses Intuition::listToText()

templates/output.html.twig
```diff
@@ -5,13 +5,19 @@
         <div class="form-group">
             <label for="image">{{ msg('image-url') }}</label>
             <input name="image" id="image" class="form-control" type="url" value="{{ app.request.get('image') }}" required />
-            <p class="help-block">{{ msg( 'image-url-help', [image_hosts] ) }}</p>
+            <p class="help-block">{{ msg( 'image-url-help', image_hosts ) }}</p>
```
samwilson: Won't this pass each host as a separate message parameter?

src/Controller/OcrController.php
```diff
@@ -108,31 +108,57 @@ public function getLangs(Request $request): array
...
+
+    /**
+     * @Route("/api/available_langs", name="apiLangs")
      * @return JsonResponse
      */
-    public function api(): JsonResponse
+    public function apiAvailableLangsAction(): JsonResponse
+    {
+        return $this->getApiResponse([
+            'available_langs' => $this->engine->getValidLangs(),
```
MusikAnimal, line 147: This consequently includes a bunch of irrelevant things in the response like the Tesseract options, and `langs` which is actually kind of confusing. I'm assuming it doesn't matter much, but if it does we can `unset` those keys just for this API endpoint. Or even return just the languages as an array, and nothing else?

samwilson: I don't think it matters too much, and might be useful if we do any normalization of language codes, e.g. requesting `/api/available_langs?langs[]=Foo-bar` returns `{available_langs:{…}, langs: ['foo_bar']}` or whatever?

Are we going to translate language names anywhere? Because then sending `?uselang=x` would change the result.

MusikAnimal: I changed it to return only `available_langs` and the request engine (since if you don't request an engine, it defaults to `google` so we should let them know that)

samwilson: :+1:

MusikAnimal: I commented before seeing your reply... I see what you're saying about language code normalization! Yes that's a good idea, but should /api/available_langs be responsible for that? Either way probably better for a separate PR?

samwilson: Yeah I think it can wait. There's a whole bunch of stuff about what the wiki-side UI is going to need that doesn't feel fully figured out yet (at least to me). :)

templates/output.html.twig
MusikAnimal: Ah, I see why we were calling listToText() in the controller now :) I've made it so it only does that for the /home route. I felt like the API should return an array but I suppose there aren't going to be many clients programmatically relying on that list.

assets/app.js
```diff
@@ -7,11 +7,11 @@ const $ = require('jquery');
 $(function () {
     // Initiate Select2, which allows dynamic entry of languages.
     $('#lang').select2({
-        tags: true,
         theme: 'bootstrap',
     });
 
     // Show engine-specific options.
+    // TODO: Update the Select2 options with available languages.
```
MusikAnimal, line 14: This I think would normally be a big problem for usability, but I understand that's less important since most people will never see the UI. I imagine we might be redesigning everything anyway, so I didn't want to put too much work into the user-facing stuff. For now you can force it to use the Tesseract list by putting ?engine=tesseract in the URL.

MusikAnimal: Also, I assume we'll want to map TesseractEngine::getValidLangs() to return ISO 639-1 for consistency with what people are used to on-wiki, and also so that Intuitions `getLangName()` will work. Right now with Tesseract's format you don't get the language name in the autocompletion.

samwilson: You do get the language name for some languages, although I'm not really sure why. I've commented on the ticket https://phabricator.wikimedia.org/T282073#7064604

PR #30, opened 2021-05-07
Fall back to default engine if given an invalid engine and show warning

MusikAnimal:
Bug: T282135

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -11,6 +11,7 @@
     "engine-help": "Choose between the open-source Wikimedia-hosted Tesseract, and Google's Cloud Vision API.",
     "engine-name-google": "Google",
     "engine-name-tesseract": "Tesseract",
+    "engine-not-found-warning": "The requested engine '$1' was not found. Using the default engine '$2' instead.",
     "submit": "Go",
     "copy-to-clipboard": "Copy to clipboard",
     "copied-to-clipboard": "Copied!",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -15,6 +15,7 @@
 	"engine-help": "Help text for the 'engine' field.",
 	"engine-name-google": "Option text for the 'google' engine.",
 	"engine-name-tesseract": "Option text for the 'tesseract' engine.",
+	"engine-not-found-warning": "Error message displayed when an invalid engine is requested.\n\nParameter:\n* $1 – the requested engine.\n* $2 – the default engine that is being used instead.",
 	"submit": "Submit button text.\n{{identical|Go}}",
 	"copy-to-clipboard": "Button text for the copy-to-clipboard button.",
 	"copied-to-clipboard": "Button text shown after the copy-to-clipboard button is clicked.",
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -6,6 +6,7 @@ namespace App\Controller;
 use App\Engine\EngineFactory;
 use App\Engine\GoogleCloudVisionEngine;
 use App\Engine\TesseractEngine;
+use App\Exception\EngineNotFoundException;
 use Krinkle\Intuition\Intuition;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 use Symfony\Component\HttpFoundation\JsonResponse;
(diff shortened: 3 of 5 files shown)


MusikAnimal:
In my opinion this seems a bit hacky... but that's what we get for putting so much logic in the controller constructor, before the container is fully built. For example, we can't simply call `$this->addFlash()` when we capture the `EngineNotFoundException`. That has to be done in a controller action, I think.

The other much, much simpler option is to silently ignore invalid engines and just use the default one. Ideally we'd also show a `warn` flash message to the user (as opposed to `danger`), but we still have the aforementioned problem of not being able to use `addFlash` in the constructor. To get around that, we could toss a bunch more logic into the `ExceptionListener`, but I think that's more convoluted than what we have in this PR.

MusikAnimal:
Another possible TODO: add tests? I started to write proper integration tests but that turned into a rabbit hole. I can put the effort in though, if we want! Or at least some simple unit testing of `OcrController::setEngine()` with `expectException()`.

i18n/en.json
```diff
@@ -11,6 +11,7 @@
     "engine-help": "Choose between the open-source Wikimedia-hosted Tesseract, and Google's Cloud Vision API.",
     "engine-name-google": "Google",
     "engine-name-tesseract": "Tesseract",
+    "engine-not-found-error": "The requested engine '$1' not found. Supported engines include: $2",
```
dayllanmaza: typo maybe?: *was not found

dayllanmaza (commented):
> This is tricky because an engine must be set for the view to be
rendered.

Maybe I missed it but why is the engine a requirement for the view? And even if it is, couldn't we just set it in the actions instead of the constructor allowing you to play with `addFlash` like you mentioned above?

i18n/en.json
MusikAnimal: Done

MusikAnimal:
> > This is tricky because an engine must be set for the view to be
> > rendered.
> 
> Maybe I missed it but why is the engine a requirement for the view? And even if it is, couldn't we just set it in the actions instead of the constructor allowing you to play with `addFlash` like you mentioned above?

output.html.twig expects the `engine` Twig variable to be set, as well as `langs` and `image_hosts` which consequently also require an engine. I tried to explain in the comments at [OcrController#L36](https://github.com/wikimedia/wikimedia-ocr/pull/30/files#diff-02887fb256410c2b4193f1f7759ddcbbf2e6b33594416c5adeae559b988816edR62) and [#L72](https://github.com/wikimedia/wikimedia-ocr/pull/30/files#diff-02887fb256410c2b4193f1f7759ddcbbf2e6b33594416c5adeae559b988816edR72). We can move all of that to private methods and call them in each action (following #29 there are three actions now), but it just makes it a bit less DRY. I suppose we could also look into using [before filters](https://symfony.com/doc/current/event_dispatcher/before_after_filters.html), which aren't as elegant in PHP/Symfony as say in Ruby on Rails, but since the Twig variables are a static property on the controller, I think it could work well in this case. Maybe I'll look into that.

There's one other option, too, which is to simply not show the form at all when given an invalid engine. Then we don't need to worry about all the other `OcrController::$params` being set. That's a bit subpar user experience, but as I understand very few people use the UI anyway.

Daimona:
First of all, I recognize that this error would only be shown if the user manually requests an invalid engine via the URL, and thus it's not very important for us to show a nice message. As such, we can proceed with the silent fallback if you prefer.

That said, I've played a bit with your proposal of initializing stuff in a private method instead of the constructor, and it worked pretty well. That method can then use addFlash and set the engine to some default for showing the page. I'm not sure if we can do this with a before filter though, because we also have non-static properties to set, most notably `$this->engine`.

Do you think that deferring initialization to a private method (that we'd have to call manually for each action) is acceptable, given the costs and benefits? As you said it's a bit less DRY, but OTOH, avoiding non-trivial tasks in the constructor is a good idea in general.

MusikAnimal:
> Do you think that deferring initialization to a private method (that we'd have to call manually for each action) is acceptable, given the costs and benefits?

Definitely acceptable, and I don't know why I was so hesitant to go that route. There are only three actions, and it's unlikely we'll add more in the future. Latest push takes everything except dependency setters out of the constructor.

Daimona:
> > Do you think that deferring initialization to a private method (that we'd have to call manually for each action) is acceptable, given the costs and benefits?
> 
> Definitely acceptable, and I don't know why I was so hesitant to go that route. There are only three actions, and it's unlikely we'll add more in the future. Latest push takes everything except dependency setters out of the constructor.

Thank you, it looks identical to what I had in mind! I understand the hesitation, the same line is being repeated three times now, but I guess it's kinda fine. Having an initialization method called directly on our Controller would be better, but the current approach works. Will review more thoroughly later.

Daimona:
I agree with the intention, the code is sane, output is escaped correctly, everything working as intended, no issues that I can think of. Nice!

PR #32, opened 2021-05-13
Cache OCR transcriptions

MusikAnimal:
1 hour by default

Bug: T282837


samwilson (commented):
This looks great!

I do wonder if it all belongs in the controller though. It feels to me like it should be something the engine class handles. For instance, if we were to add a CLI entry point, it'd not have access to the caching (not that we're going to of course, so probably this is totally fine!).

Also, I think we'll need to add something to clear the stale items. I guess that's just a separate cronjob though?

PR #33, opened 2021-05-27
Add language map and refactor EngineBase::getValidLangs()

MusikAnimal:
This also adds the extra non-standard languages offered by Tesseract,
along with their localized names.
    
The Select2 control on the form now dynamically updates with the
supported languages when the engine is changed.
    
Bug: https://phabricator.wikimedia.org/T282760
Bug: https://phabricator.wikimedia.org/T281866
Bug: https://phabricator.wikimedia.org/T282073

### DIFF
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -17,12 +17,48 @@ abstract class EngineBase
     /** @var Intuition */
     protected $intuition;
 
-    public function __construct(Intuition $intuition)
+    /** @var string */
+    protected $projectDir;
+
+    /** @var string[][] Local PHP array copy of langs.json */
+    protected $langList;
+
+    /** @var string[] Additional localized names for non-standard language codes. */
+    public const LANG_NAMES = [
+        'az-cyrl' => 'Azərbaycan (qədim yazı)',
+        'de-frk' => 'Deutsch (Fraktur)',
+        'enm' => 'Middle English (1100-1500)',
+        'es-old' => 'español (viejo)',
+        'equ' => 'Math / equation detection module',
+        'fro' => 'Franceis, François, Romanz (1400-1600)',
+        'it-old' => 'italiano (vecchio)',
+        'ka-old' => 'ქართული (ძველი)',
+        'ko-vert' => '한국어 (세로)',
+        'osd' => 'Orientation and script detection module',
+        'sr-latn' => 'Српски (латиница)',
+        'syr' => 'leššānā Suryāyā',
+        'uz-cyrl' => 'oʻzbekcha',
+    ];
+
+    /**
+     * EngineBase constructor.
+     * @param Intuition $intuition
+     * @param string $projectDir
+     */
+    public function __construct(Intuition $intuition, string $projectDir)
     {
         $this->intuition = $intuition;
+        $this->projectDir = $projectDir;
     }
 
     /**
+     * Unique identifier for the engine.
+     * @return string
+     */
+    abstract public static function getId(): string;
+
+    /**
+     * Get transcribed text from the given image.
      * @param string $imageUrl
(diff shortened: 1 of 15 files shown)


src/Twig/AppExtension.php
```diff
@@ -28,4 +53,14 @@ public function getTextareaRows(string $text): int
     {
         return max(10, substr_count($text, "\n"));
     }
+
+    /**
+     * Get the name of the given language. This adds a few translations that don't exist in Intuition.
+     * @param string|null $lang
+     * @return string
+     */
+    public function getOcrLangName(?string $lang = null): string
```
MusikAnimal, line 62: FYI, the original idea was to extend ToolforeBundle's `Extension` class and override the `getLangName` method. However we can't do this because ToolforgeBundle type-hints `Session` rather than `SessionInterface` which confuses Symfony in this context (we'd have to specify the DI in services.yml). I can still do that if we feel it's better.

MusikAnimal:
See info on how I built the langs.json list at https://phabricator.wikimedia.org/T282760#7130811

.env.test
```diff
@@ -0,0 +1 @@
+KERNEL_CLASS=App\Kernel
```
samwilson: This could be replaced by `<server name="KERNEL_CLASS" value="App\Kernel" />` in `phpunit.xml.dist`, to save the addition of a new file and keep all the test env vars together.

src/Engine/EngineBase.php
```diff
@@ -17,29 +18,126 @@ abstract class EngineBase
...
     abstract public function getText(string $imageUrl, ?array $langs = null): string;
 
     /**
+     * Get the language list from langs.json
+     * @return string[][]
+     */
+    private function getLangList(): array
+    {
+        if (!$this->langList) {
+            $this->langList = json_decode(file_get_contents($this->kernel->getProjectDir().'/public/langs.json'), true);
```
samwilson: This is the only use of $kernel isn't it? Would it be better to inject the project directory only, rather than the whole kernel? e.g. in services.yaml:

```lang=yaml
    App\Engine\EngineBase:
        arguments:
            $projectDir: '%kernel.project_dir%'
```

samwilson (changes requested):
This looks great, and works almost perfectly. The only bug I can find wrong is that it's not respecting pre-set langs from the URL, e.g. `http://localhost:8000/?langs[]=it&langs[]=en` doesn't prefill the languages.

The other thing I was wondering, and you've probably already dug into it and concluded that it's not simple, is that Select2 supports fetching its own JSON source (with the `ajax` config key), as well as munging the results from that call into the `id`/`text` structure required. That all works fine, but I guess gets confusing when it needs to be reset by changes in the engine form field?

.env.test
MusikAnimal: Done!

src/Engine/EngineBase.php
MusikAnimal: Good call! Done.

MusikAnimal:
> The other thing I was wondering, and you've probably already dug into it and concluded that it's not simple, is that Select2 supports fetching its own JSON source (with the ajax config key), as well as munging the results from that call into the id/text structure required. That all works fine, but I guess gets confusing when it needs to be reset by changes in the engine form field?

We could use that, but then it queries on every keystroke (I think?), when we really only need to make a request when the engine changes. It's a fast API endpoint so performance probably wouldn't be an issue, but it just felt wrong to me in this case.

samwilson:
> We could use that, but then it queries on every keystroke (I think?), when we really only need to make a request when the engine changes. It's a fast API endpoint so performance probably wouldn't be an issue, but it just felt wrong to me in this case.

Good point! I hadn't thought of that.

This all looks good to me.

PR #35, opened 2021-06-08
Catch all internal exceptions from tesseract

Daimona:
For now, this uses a generic error message as it's not easy to be
informative without risking to output sensitive system info.

Bug: T282141

### DIFF
--- a/src/EventListener/ExceptionListener.php
+++ b/src/EventListener/ExceptionListener.php
@@ -6,6 +6,7 @@ namespace App\EventListener;
 use App\Controller\OcrController;
 use App\Exception\OcrException;
 use Krinkle\Intuition\Intuition;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
@@ -13,7 +14,7 @@ use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
-use thiagoalessio\TesseractOCR\TesseractOcrException;
+use thiagoalessio\TesseractOCR\UnsuccessfulCommandException;
 use Twig\Environment;
 
 class ExceptionListener
@@ -30,24 +31,29 @@ class ExceptionListener
     /** @var Intuition */
     private $intuition;
 
+    /** @var LoggerInterface */
+    private $tesseractLogger;
+
     public function __construct(
         RequestStack $requestStack,
         SessionInterface $session,
         Environment $twig,
-        Intuition $intuition
+        Intuition $intuition,
+        LoggerInterface $tesseractLogger
     ) {
         $this->request = $requestStack->getCurrentRequest();
         $this->session = $session;
         $this->twig = $twig;
         $this->intuition = $intuition;
+        $this->tesseractLogger = $tesseractLogger;
     }
 
     public function onKernelException(ExceptionEvent $event): void
     {
         $exception = $event->getThrowable();
 
-        // We only care about OcrExceptions, and TesseractOcrException thrown by the library (T282141).
-        if (!( $exception instanceof OcrException || $exception instanceof TesseractOcrException )
+        // We only care about OcrExceptions, and UnsuccessfulCommandException thrown by the library (T282141).
+        if (!( $exception instanceof OcrException || $exception instanceof UnsuccessfulCommandException )
             || !$event->isMasterRequest()
         ) {
             return;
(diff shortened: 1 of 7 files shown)


src/EventListener/ExceptionListener.php
```diff
@@ -45,8 +46,10 @@ public function onKernelException(ExceptionEvent $event): void
     {
         $exception = $event->getThrowable();
 
-        // We only care about OcrExceptions.
-        if (!$exception instanceof OcrException || !$event->isMasterRequest()) {
+        // We only care about OcrExceptions, and TesseractOcrException thrown by the library (T282141).
+        if (!( $exception instanceof OcrException || $exception instanceof TesseractOcrException )
```
MusikAnimal: The only issue here (I think) is that this will suppress error emails. That's fine if it's due to user-error, but for instance if `NoWritePermissionsForOutputFile` happens (which extends `TesseractOcrException`), we would never know about it until someone tells us. In particular `UnsuccessfulCommandException` could be any number of things that only we can fix, so I think it's preferred we get notified about them.

We can still customize the 500 error pages to be more friendly. If you were to capture `TesseractOcrException`s and re-`throw` a generic `Exception` with a customized message, we ensure this message is shown by making our own custom 500 error page: https://symfony.com/doc/current/controller/error_pages.html

Daimona: Fair point! Alternatively, would it be possible to log the exception somewhere so that an alert email is sent out, but respond with a 200 regardless? I'm not sure how email notifications are configured, so it might well be impossible.

If it's impossible, customizing the 500 error page seems a good idea. Perhaps we might also want to limit this to `UnsuccessfulCommandException` for now, since other exceptions seems serious enough to warrant a 500 (e.g. `TesseractNotFoundException`)

MusikAnimal: > would it be possible to log the exception somewhere so that an alert email is sent out, but respond with a 200 regardless?

Probably, but I don't know that there's any out-of-the-box solution for this that will respect the [Monolog configuration](https://github.com/wikimedia/wikimedia-ocr/blob/main/config/packages/prod/monolog.yaml). I'm sure there's a way to do it, though!

So customizing the 500 error page may be the best route. I also agree we can limit this to `UnsuccessfulCommandException`. For what we need, I think all we have to do is [override the default error template](https://symfony.com/doc/current/controller/error_pages.html#overriding-the-default-error-templates). You can add `{% extends 'base.html.twig' %}` and put the content inside a `{% block body %}` to maintain the same design as the rest of the app, just like we do in [output.html.twig](https://github.com/wikimedia/wikimedia-ocr/blob/main/templates/output.html.twig) (I say this in case you aren't familiar with Twig or Symfony). There you also have access to the Exception object and can show the message with i.e. `{{ exception.message }}`.

MusikAnimal: >> would it be possible to log the exception somewhere so that an alert email is sent out, but respond with a 200 regardless?

> Probably, but I don't know that there's any out-of-the-box solution for this that will respect the Monolog configuration. I'm sure there's a way to do it, though!

After reviewing [the docs](https://symfony.com/doc/current/logging/monolog_email.html), this might be as simple as adding a new channel to the monolog configuration. Then I suppose you could inject the Monolog service into this class then call `$this->logger->error()` or (`crit()`) and that would tell Symfony to send an email? I have never tried this.

For monolog.yaml:
```yaml
monolog:
    channels: ['main', 'tesseract']
    handlers:
        main:
            …
        tesseract:
            type: fingers_crossed
            action_level: error
            handler: nested
        …
```

Then in services.yml:

```yaml
services:
    …
    App\Exception\OcrException:
        arguments:
            $logger: 'monolog.logger.tesseract'
```

then change `OcrException:__construct()` to take a `LoggerInterface $logger`. Set it as a property and _maybe_, just maybe something like this will work: `$this->logger->error($this->getMessageForTesseractException($exception))`.

The monolog.yaml system I find pretty confusing, but something like the above might work. If we can get that working, that'd be more ideal I think that customizing an error page.

Note also you should be able to test things locally by changing your `APP_LOG_RECIPIENT_1` in .env.local, and temporarily making the above changes to config/packages/dev/monolog.yaml. But if you need to, feel free to test the change to /prod/monolog.yaml on our [staging server](https://openstack-browser.toolforge.org/server/wsexport-dev01.wikisource.eqiad1.wikimedia.cloud), which you should have access to. Let me know if you need help with that.

Daimona: Nice, thank you, that's exactly what I was thinking of, except for how to do it in practice :) I'll play with it later today and will report back once I have something to share.

Daimona: So, this is now working fine locally, but I don't understand the CI failure.

Daimona: @MusikAnimal I can't figure out what's wrong; this is still working for me locally, where I get:

```
$ php bin/console debug:container | grep "monolog\.logger\.tesseract"
  Psr\Log\LoggerInterface $tesseractLogger                                     alias for "monolog.logger.tesseract"
  monolog.logger.tesseract                                                     Symfony\Bridge\Monolog\Logger
```

If I'm reading the [docs](https://symfony.com/doc/current/logging/channels_handlers.html#configure-additional-channels-without-tagged-services) correctly, the new service should be auto-created.

I've also made sure to push everything. Does it work for you?

MusikAnimal: It errored out on my local. Looks like we need to also define this monolog channel for the other environments.

For dev/monolog.yaml I tried:
```yaml
monolog:
    channels: ['main', 'console', 'tesseract']
    handlers:
        main:
            …
        tesseract:
            type: fingers_crossed
            action_level: error
            handler: console
        console:
            …
```

and for test/monolog I used:
```yaml
monolog:
    channels: ['main', 'nested', 'tesseract']
    handlers:
        main:
            …
        tesseract:
            type: fingers_crossed
            action_level: error
            handler: nested
        nested:
            …
```

and that seemed to work for me.

Daimona: > It errored out on my local. Looks like we need to also define this monolog channel for the other environments.

Dang, I missed that part, thank you! Indeed, I was testing this locally with the prod env; switching back to dev results in the same error. Fixing in a minute.

MusikAnimal (approved):
I tested this on staging an email was sent. Merging!

PR #38, opened 2021-06-09
Strengthen the configuration options for phan

Daimona:
- Add phan-taint-check-plugin to spot possible security issues (although it still cannot understand templates)
- Pass --long-progress-bar explicitly to make sure the progress bar it's always shown
- Add a few config options for stricter analysis. This is a good set of options, and doesn't cause any new issue to be reported
- Install the php-ast extension in the github CI to run phan faster

Note that we need to downgrade phan due to the strict dependency that taint-check has.

Bug: T284571

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,8 +29,8 @@
     "require-dev": {
         "drenso/phan-extensions": "^3.3",
         "mediawiki/minus-x": "^1.1",
-        "phan/phan": "3.2.6",
         "mediawiki/phan-taint-check-plugin": "3.2.1",
+        "phan/phan": "3.2.6",
         "slevomat/coding-standard": "^4.8",
         "symfony/phpunit-bridge": "^5.2",
         "symfony/stopwatch": "^5.2",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -77,7 +78,7 @@
             "./bin/console lint:twig ./templates",
             "./bin/console lint:yaml ./config",
             "minus-x check .",
-            "phan --allow-polyfill-parser",
+            "phan --allow-polyfill-parser --long-progress-bar",
```
samwilson, line 81: Why add this? (Not that it matters, seems sensible; just wondering!)

samwilson (approved):
Looks great! Thanks for improving this.

composer.json
Daimona: I was originally under the impression that the default would be to use no progress bar at all. Then I realized that `long-progress-bar` is actually the default, but I figured I would leave it anyway so it's more explicit.

As for why this specific choice: I think a progress bar is always necessary for this kind of tools, since they might become slow-ish, and the other progress bar (that you'd get with `-p`) is not really suitable for CI logs because it takes up a lot of space.

PR #41, opened 2021-06-10
Add numbers to allowed chars in lang codes

samwilson:


Daimona:
Formally approving your change, you can self-merge if you approve mine.

PR #43, opened 2021-06-14
EngineBase: add missing lang name for 'frm'

MusikAnimal:
Also remove 'equ' as it's been removed from langs.json

Bug: T282760

### DIFF
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -29,7 +29,7 @@ abstract class EngineBase
         'de-frk' => 'Deutsch (Fraktur)',
         'enm' => 'Middle English (1100-1500)',
         'es-old' => 'español (viejo)',
-        'equ' => 'Math / equation detection module',
+        'frm' => 'moyen français (1400-1600)',
         'fro' => 'Franceis, François, Romanz (1400-1600)',
         'it-old' => 'italiano antico',
         'ka-old' => 'ქართული (ძველი)',


src/Engine/EngineBase.php
```diff
@@ -29,7 +29,7 @@ abstract class EngineBase
         'de-frk' => 'Deutsch (Fraktur)',
         'enm' => 'Middle English (1100-1500)',
         'es-old' => 'español (viejo)',
-        'equ' => 'Math / equation detection module',
+        'frm' => 'français moyen (1400-1600)',
```
Daimona: I think this should be "moyen français", according to https://fr.wikipedia.org/wiki/Moyen_fran%C3%A7ais

Daimona: I mean, "moyen français (1400-1600)", the swap is the important thing.

MusikAnimal: Done

Daimona:
Thank you! Tomorrow I might follow-up with a test that ensures that every language has an entry where it needs to.

MusikAnimal:
> Thank you! Tomorrow I might follow-up with a test that ensures that every language has an entry where it needs to.

Good idea! I just submitted #44

Daimona:
> > Thank you! Tomorrow I might follow-up with a test that ensures that every language has an entry where it needs to.
> 
> Good idea! I just submitted #44

Thank you!

PR #45, opened 2021-06-19
Add bash script to validate tesseract

Daimona:
This script validates the current tesseract installation upon composer install/update events (and `composer test`, mostly useful if you change something with your tesseract installation). This should help with T284831 and T282150. As noted at T284831, we're assuming that bash is available (i.e. a UNIX-like environment).


samwilson (approved):
Looks good to me!

PR #49, opened 2021-06-25
Ignore invalid languages for API requests

Daimona:
**NEW:**
Instead of failing hard with an exception, we add a warning to the
response.
This is mainly useful for requests from wikisource, where we pass the
content language by default, but it might be unsupported.

Also make the engines return structured output and add code for
supporting warnings even in the UI, although it's currently impossible
to get any.

Bug: T285544

**OLD:**
This is only useful for requests from wikisource, where we pass the
content language by default, but it might be unsupported. As such, this
option is not exposed in the UI.

Bug: T285544

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -3,7 +3,9 @@ declare(strict_types = 1);
 
 namespace App\Controller;
 
+use App\Engine\EngineBase;
 use App\Engine\EngineFactory;
+use App\Engine\EngineResult;
 use App\Engine\GoogleCloudVisionEngine;
 use App\Engine\TesseractEngine;
 use App\Exception\EngineNotFoundException;
@@ -158,7 +160,11 @@ class OcrController extends AbstractController
         static::$params['image_hosts'] = $this->intuition->listToText(static::$params['image_hosts']);
 
         if ($this->imageUrl) {
-            static::$params['text'] = $this->getText();
+            $result = $this->getResult(EngineBase::ERROR_ON_INVALID_LANGS);
+            static::$params['text'] = $result->getText();
+            foreach ($result->getWarnings() as $warning) {
+                $this->addFlash('warning', $warning);
+            }
         }
 
         return $this->render('output.html.twig', static::$params);
@@ -172,9 +178,15 @@ class OcrController extends AbstractController
     public function apiAction(): JsonResponse
     {
         $this->setup();
-        return $this->getApiResponse(array_merge(static::$params, [
-            'text' => $this->getText(),
-        ]));
+        $result = $this->getResult(EngineBase::WARN_ON_INVALID_LANGS);
+        $responseParams = array_merge(static::$params, [
+            'text' => $result->getText(),
+        ]);
+        $warnings = $result->getWarnings();
+        if ($warnings) {
+            $responseParams['warnings'] = $warnings;
+        }
+        return $this->getApiResponse($responseParams);
     }
 
     /**
@@ -218,9 +230,10 @@ class OcrController extends AbstractController
 
     /**
      * Get and cache the transcription based on options set in static::$params.
-     * @return string
+     * @param string $invalidLangsMode EngineBase::WARN_ON_INVALID_LANGS or EngineBase::ERROR_ON_INVALID_LANGS
+     * @return EngineResult
      */
-    private function getText(): string
+    private function getResult(string $invalidLangsMode): EngineResult
     {
         $cacheKey = md5(implode(
             '|',
@@ -229,12 +242,14 @@ class OcrController extends AbstractController
                 static::$params['engine'],
                 implode('|', static::$params['langs']),
                 static::$params['psm'],
+                // Warning messages are localized
+                $this->intuition->getLang(),
             ]
         ));
 
-        return $this->cache->get($cacheKey, function (ItemInterface $item) {
+        return $this->cache->get($cacheKey, function (ItemInterface $item) use ($invalidLangsMode) {
             $item->expiresAfter((int)$this->getParameter('cache_ttl'));
-            return $this->engine->getText($this->imageUrl, static::$params['langs']);
+            return $this->engine->getResult($this->imageUrl, $invalidLangsMode, static::$params['langs']);
         });
     }
 }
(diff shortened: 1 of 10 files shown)


dayllanmaza:
I'm not sure this is the way to go. Can't we prevent the wikis from passing invalid languages? The whole point of validating languages is that we don't set false expectations about what is supported and what is not

Daimona:
> I'm not sure this is the way to go. Can't we prevent the wikis from passing invalid languages? The whole point of validating languages is that we don't set false expectations about what is supported and what is not

I think the alternative would be to send a request to the available_langs endpoint, and excluding the `langs` parameter if the content language is not listed. The point is that the language selection is completely transparent to the user. They don't know that the content language will be sent with the request, nor is it something they can control. OTOH, querying that endpoint would add an additional request to the OCR tool for every first transcription on a given page.

This option is also meant to be used for that kind of requests only (i.e. from the wiki, upon clicking the OCR button), so not via the UI, and ideally not when manually querying the API endpoint.

Of course, things would be different if we allowed the user to specify one or more languages. But even then, I think we might be fine with keeping the behaviour from this PR, as long as we also show a warning once the request completes (e.g. "The following languages aren't supported and were ignored: X, Y, Z".

samwilson:
I'm a bit confused about this too, sorry!

I get the idea of `?ignore-invalid-langs=1`, but I'm not sure when we would want to *not* pass that. i.e. I assume we'll make the wiki code send this param, and so for almost all requests we'll fall back to no language (and leave it to the engine to do its best). Or is this more about people manually specifying multiple languages, some of which might be invalid, and so sometimes they'd want to be told about that and sometimes not told?

It seems like we might more simply just drop invalid language codes, perhaps also adding a message in the response that we've done so, but not making it necessary to opt-in to doing so.

The primary use (if I'm understanding things) is that we don't want to fail wikis in languages that are not supported, and for those we're happy to fall back to not specifying a language. Is that correct?

Daimona:
> I get the idea of `?ignore-invalid-langs=1`, but I'm not sure when we would want to _not_ pass that.

I think it should not be passed for all non-API requests (i.e. via the UI). For API requests, I think it depends on the client. For wikisource, we'd always be using it.

> i.e. I assume we'll make the wiki code send this param, and so for almost all requests we'll fall back to no language (and leave it to the engine to do its best).

I don't know how many requests will end up using an unsupported language but yeah, that's the idea.

> Or is this more about people manually specifying multiple languages, some of which might be invalid, and so sometimes they'd want to be told about that and sometimes not told?

I think if we add an option for the users to manually specify languages on-wiki, we'd always want to show a warning if any of the provided languages is unsupported. But that's just because at that point, the process of sending language hints would no longer be transparent to the user, so they should receive feedback. And assuming that this language selection option will be implemented via a UI element (as opposed to e.g. some site-wide configuration), I don't think we should ever fallback silently in that case.

> It seems like we might more simply just drop invalid language codes, perhaps also adding a message in the response that we've done so, but not making it necessary to opt-in to doing so.

I think this might make sense, given that for API requests, our only use-case is with this option set, as I said above. If we also add a warning in the response to inform the client that some languages were invalid, I think this could also work for the UI. Let's just do that?

> The primary use (if I'm understanding things) is that we don't want to fail wikis in languages that are not supported, and for those we're happy to fall back to not specifying a language. Is that correct?

Yep.

samwilson:
> > I get the idea of `?ignore-invalid-langs=1`, but I'm not sure when we would want to _not_ pass that.
> 
> I think it should not be passed for all non-API requests (i.e. via the UI). For API requests, I think it depends on the client. For wikisource, we'd always be using it.
> 

That makes sense. For the UI, it shouldn't be possible to pass invalid langs anyway, because they're fixed. Someone could change the URL of course, but if they do then it'd be better to show a warning.

Given that we're building the only client that's using this at the moment, I don't think it makes sense to add complexity that we don't need yet.

> 
> > It seems like we might more simply just drop invalid language codes, perhaps also adding a message in the response that we've done so, but not making it necessary to opt-in to doing so.
> 
> I think this might make sense, given that for API requests, our only use-case is with this option set, as I said above. If we also add a warning in the response to inform the client that some languages were invalid, I think this could also work for the UI. Let's just do that?
> 

Sounds good to me.

Daimona:
Done, and also generalized the code a bit to have the engine return structured information, and allow warnings to be displayed in the UI (although it's currently impossible to get any).

src/Controller/OcrController.php
```diff
@@ -216,9 +229,9 @@ private function getText(): string
             ]
```
samwilson, line 247: Now that the result includes translated messages, I think we need the interface language to be part of the cache key, otherwise the warning messages won't be translated for subsequent requests. e.g. the following should return different results, but are both in English:

```bash
{pr/49} ~/code/wikimedia/wikimedia-ocr$ curl -s 'http://localhost:8000/api?image=http%3A%2F%2Flocalhost%2Fwikimedia%2Fwiki1%2Fimages_wiki1%2Fthumb%2Fa%2Faf%2FTestdoc.pdf%2Fpage2-1024px-Testdoc.pdf.jpg&engine=tesseract&psm=3&lang=asdfasd&uselang=en' | jq .warnings
[
  "The following languages are invalid or not supported by the engine and were ignored: asdfasd"
]
{pr/49} ~/code/wikimedia/wikimedia-ocr$ curl -s 'http://localhost:8000/api?image=http%3A%2F%2Flocalhost%2Fwikimedia%2Fwiki1%2Fimages_wiki1%2Fthumb%2Fa%2Faf%2FTestdoc.pdf%2Fpage2-1024px-Testdoc.pdf.jpg&engine=tesseract&psm=3&lang=asdfasd&uselang=qqx' | jq .warnings
[
  "The following languages are invalid or not supported by the engine and were ignored: asdfasd"
]
```

It might be better to add the warnings after retrieving the cached data, to avoid extra cache items (although I don't really think it's likely to be that much of a problem: we probably don't serve the same page in different interface languages all that often, especially with invalid language codes).

samwilson:
Sorry for being slow to look at this again, @Daimona!

src/Controller/OcrController.php
Daimona: Good point, done! I went with adding the interface language to the cache key. An alternative might be to store msg key + params in EngineResult->warnings, and then localize them in the OcrController, which is I think what you meant by "add the warnings after retrieving the cached data". I think that would require a few changes though, e.g. for proper handling of array parameters, which should be run through listToText etc., and I'm also unsure if we're going to need that level of precision. Do you think the current solution (i.e. varying on the interface language) is enough for now?

samwilson: Yep, keeps it simple! Looks good to me.

samwilson (approved):

PR #50, opened 2021-06-25
Polish the UI, add logo

MusikAnimal:
Bug: T283879


MusikAnimal:
FYI I showed some screenshots to @nayoub and he signed off on them. What you see may not match the Figma designs exactly.

samwilson:
Looks great!

The only slight oddity I noticed was that when there's an error message, the warning box is much wider than the form. Is this correct? e.g.:

ocr-pr50

MusikAnimal:
> The only slight oddity I noticed was that when there's an error message, the warning box is much wider than the form. Is this correct?

Good question! That was not in the design spec. I think it looks okay full-width, since it matches the width of the footer and the output (you can see error message + output from entering an invalid engine, for example), but I think it would also look good being the same width as the form elements. I'm leaning more towards the latter. @nayoub what do you think?

i18n/qqq.json
```diff
@@ -24,9 +24,11 @@
 	"documentation": "Link text in the footer for the documentation link\n{{identical|Documentation}}",
 	"version": "Link text in the footer for the source code link.\n\nParameter:\n* $1 – the current application version number.\n\n{{identical|Version}}",
 	"report-issue": "Link text in the footer for the issue-reporting link.",
+	"langs-placeholder": "Placeholder text for the language input field.",
 	"langs-param-error": "Error message displayed when invalid language(s) are submitted.\n\nParameters:\n* $1 – number of invalid languages\n* $2 - the list of invalid languages",
 	"tesseract-options": "Heading for Tesseract-specific options.",
 	"tesseract-psm-label": "Form label for the Tesseract page segmentation mode.",
+	"tesseract-psm-help": "Help text for the Tesseract page segmentation mode option. 'Sparse text' refers to options, see messages {{msg-wm|tesseract-psm-11}} and {{msg-wm|tesseract-psm-12}}.",
```
Daimona, line 31: The way I've seen it done on many MW repos is

  {{doc-important|Do not translate \"Sparse text\".}}

can we use it here?

templates/output.html.twig
```diff
@@ -2,46 +2,56 @@
...
-        </div>
+        <fieldset>
+            <legend>{{ msg('form-heading') }}</legend>
+            <div class="form-group">
+                <label for="image">{{ msg('image-url') }}</label>
+                <input name="image" id="image" class="form-control" type="url" value="{{ app.request.get('image') }}" required />
+                <p class="help-block">{{ msg( 'image-url-help', [image_hosts] ) }}</p>
+            </div>
+            <div class="form-group">
+                <label for="engine">{{ msg('engine') }}</label>
```
Daimona: This is invalid HTML, since the "for" attribute no longer has a valid element ID. Since both radios already have labels, this "for" attribute can just be removed.

assets/styles/app.css
```diff
@@ -3,3 +3,74 @@
...
+	margin: 25px 0;
+}
+
+fieldset {
+	max-width: 541px;
+}
+
+/* Avoid select2 input from exceeding viewport on smaller screens */
+.select2-container {
+	/* stylelint-disable declaration-no-important */
```
Daimona, line 56: (I was going to say that we might've avoided this by overloading the selectors, but then I realized that the width is set with inline styling. So yeah, we need to keep it)

Daimona (commented):
I also have a general question. The tesseract options have their own fieldset, so it looks like that part of the form is completely separated from the rest. Especially because the text size etc. is identical to that of the "Transcribe an image" label, which I suppose should be the label for the whole form, including tesseract options. Can we reduce the size of the tesseract options label or something like that? I'm also not 100% sure about the engine selection and tesseract options being separated by the language selection (i.e. I'd expect the tesseract options to appear directly beneath the engine selection). But since this was changed from the previous layout, I guess it might be intentional?

i18n/qqq.json
MusikAnimal: We _do_ want them to translate it, but it should match the wording used in the 'tesseract-psm-11' and 'tesseract-psm-12' messages. I probably could have worded this better

Daimona: Oh, I'm sorry, I didn't get it.

MusikAnimal:
> I also have a general question. The tesseract options have their own fieldset, so it looks like that part of the form is completely separated from the rest. Especially because the text size etc. is identical to that of the "Transcribe an image" label, which I suppose should be the label for the whole form, including tesseract options. Can we reduce the size of the tesseract options label or something like that? I'm also not 100% sure about the engine selection and tesseract options being separated by the language selection (i.e. I'd expect the tesseract options to appear directly beneath the engine selection). But since this was changed from the previous layout, I guess it might be intentional?

Yes, all of this is going off of the [design spec](https://www.figma.com/file/vmaNXY952auqX6nw2PCh0b/Wikisource%2FOCR?node-id=302%3A0) (there are some discrepancies, but as I [mentioned above](https://github.com/wikimedia/wikimedia-ocr/pull/50#issuecomment-868774885), Nicolas signed off on it before I created the PR).

templates/output.html.twig
MusikAnimal: Done!

nayoub:
Apologies for missing this @samwilson @MusikAnimal @Daimona 
I agree with you @MusikAnimal for the error message, it would ideally be the same width as the form elements as they might appear without the output which might make it look odd.

nayoub:
@Daimona I agree with you. The "Tesseract options" could have been displayed directly after the engine selection but given the previous structure of the page, I thought it would be easier to implement that way.

Daimona:
Ok, thank you. I will take a final look once the errorbox width is decreased.

MusikAnimal:
> for the error message, it would ideally be the same width as the form elements as they might appear without the output which might make it look odd.

Done!

assets/styles/app.css
```diff
@@ -3,3 +3,75 @@
...
+.page-header .container {
+	align-items: center;
+	display: flex;
+	/* Width of container + logo size and it's padding */
+	max-width: calc( 1170px + ( (50px + 20px) * 2) );
+	width: auto;
+}
+
+.logo {
+	float: left;
```
samwilson: The logo is already a flex item, so shouldn't need to float. I think instead it needs no top and bottom margin, and only a right margin (to keep it away from the title text). At the moment it doesn't quite line up:

ocr logo alignment

without float and with the margin fixed it's like:

ocr logo alignment fixed

...oh, hang on, sorry: i'm totally wrong, the way you've got it matches the designs! oops. Well, anyway, it probably still doesn't need the float...

i18n/en.json
```diff
@@ -1,28 +1,30 @@
...
+    "image-url-help": "Insert an image URL hosted on a Wikimedia server such as: $1",
     "image-url-error": "Image URL must begin with {{PLURAL:$1|the following domain name|one of the following domain names}} and end with a valid file extension: $2",
     "image-alt-text": "The original image",
-    "language-code": "Two-letter language code (optional):",
-    "language-code-help": "The ISO639 code of the language of the text in the image.",
-    "engine": "OCR engine:",
-    "engine-help": "Choose between the open-source Wikimedia-hosted Tesseract, and Google's Cloud Vision API.",
-    "engine-name-google": "Google",
-    "engine-name-tesseract": "Tesseract",
+    "language-code": "Language (optional)",
```
Daimona: Since multiple languages are accepted, should this be "Languages (optional)"?

Daimona:
> @Daimona I agree with you. The "Tesseract options" could have been displayed directly after the engine selection but given the previous structure of the page, I thought it would be easier to implement that way.

@nayoub  I think the current position is fine. My (nit-picky) concern was more about the two fieldsets ("Transcribe an image" and "Tesseract options") having the same size. Also, strictly speaking, the former doesn't seem the label for a fieldset, but rather for the whole form. Do you think it would make sense to change the font size for one of them? We could either make the first heading bigger (e.g. by using a `<h1>` element), or the second one smaller (e.g. 18px instead of 21px). Or we can just leave it as is, if you think it makes sense :) 

Anyway, besides the i18n-related question and this fieldset question, LGTM overall.

i18n/en.json
MusikAnimal: Done!

assets/styles/app.css
MusikAnimal: Yeah I was going off of the designs. I've removed the `float: left`, though (which I got from WS Export).

samwilson (approved):
This is ready to merge I think, once the header-size question for @nayoub has been resolved (although, we could also do that as a follow-up).

Daimona (approved):
WFM, and I agree that the headers thingy can be done in a follow-up. Will merge this in a day or two if nobody objects or beats me to it.

PR #51, opened 2021-06-25
Expose flash messages in API response

MusikAnimal:
Bug: T282135

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -13,6 +13,8 @@ use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBag;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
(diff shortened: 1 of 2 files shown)


MusikAnimal:
See https://phabricator.wikimedia.org/T282135#7174690 for context

src/Controller/OcrController.php
```diff
@@ -193,6 +201,15 @@ private function getApiResponse(array $params): JsonResponse
         $response->setEncodingOptions(JSON_NUMERIC_CHECK);
         $response->setStatusCode(Response::HTTP_OK);
 
+        // Expose flash messages.
+        /** @var FlashBag $flashBag */
+        $flashBag = $this->session->getBag('flashes');
+        // @phan-suppress-next-line PhanUndeclaredMethod
```
Daimona: (FTR, you can avoid the double suppression by adding

```
'@phan-var FlashBag $flashBag';
```

at line 207, as a string literal).

MusikAnimal: Nice! I did not know about that, done. A pity that Phan can't read comments!

Daimona: Yeah, unfortunately that's a hard limitation. Inline comments are stripped by the php-ast extension when creating the AST, so phan doesn't even know that they exist.

PR #53, opened 2021-08-05
Move textarea height to CSS only

samwilson:
Rather than using JS to resize the textarea (which can fail if
the image takes too long to load), do it in CSS with flexbox.

Bug: T288190

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -64,9 +64,4 @@ $(function () {
             $copyButton.text($copyButton.data('copied-text'));
         });
     }
-
-    // Make textarea match height of image.
-    $('#text').css({
-        height: $('#source-image').outerHeight(),
-    });
 });
--- a/assets/styles/app.css
+++ b/assets/styles/app.css
@@ -69,8 +69,29 @@ fieldset,
 	margin-top: 40px;
 }
 
+.image-row-outer {
+	position: relative;
+}
+
 .copy-button {
 	position: absolute;
+	/* Relative to .image-row-outer above. */
 	right: 30px;
 	top: 15px;
+}
+
+.image-row {
+	display: flex;
+}
+
+.image-row .image {
+	flex: 1;
+	margin-right: 0.5em;
+}
+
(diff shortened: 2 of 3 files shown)


Daimona (commented):
Tested, looks good. Just one question: if I resize the window (or just view this from mobile), the textbox is actually bigger than the image. Is this expected? I'd say it's acceptable anyway.

samwilson:
> Tested, looks good. Just one question: if I resize the window (or just view this from mobile), the textbox is actually bigger than the image. Is this expected? I'd say it's acceptable anyway.

Good point, thanks. You're right, it's no longer nicely responsive.

Actually, I think this change might not be a good idea after all. I was making it because the height-finding was getting even more flaky with the addition of Cropper.js in #53 (because it hides the image, so depending on load order/speed, the textarea was appearing too small quite often), but since then I learnt of the `Cropper.ready()` function, which I think is a [better place](https://github.com/wikimedia/wikimedia-ocr/pull/54/files#diff-489e15dd56f2315dc02d3d1723571cdad5a1b6848609eb5aeb0b476118bb8236R85) to do the height-matching.

Daimona:
> > Tested, looks good. Just one question: if I resize the window (or just view this from mobile), the textbox is actually bigger than the image. Is this expected? I'd say it's acceptable anyway.
> 
> Good point, thanks. You're right, it's no longer nicely responsive.
> 
> Actually, I think this change might not be a good idea after all. I was making it because the height-finding was getting even more flaky with the addition of Cropper.js in #53 (because it hides the image, so depending on load order/speed, the textarea was appearing too small quite often), but since then I learnt of the `Cropper.ready()` function, which I think is a [better place](https://github.com/wikimedia/wikimedia-ocr/pull/54/files#diff-489e15dd56f2315dc02d3d1723571cdad5a1b6848609eb5aeb0b476118bb8236R85) to do the height-matching.

Roger roger :) I agree with that, thank you!

PR #54, opened 2021-08-05
Add image cropping

samwilson:
Add ability to crop an image before OCR,
and the Cropper.js library for selecting the image area.

Bug: https://phabricator.wikimedia.org/T288190
Bug: https://phabricator.wikimedia.org/T269818


src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @param int[] $crop
+     */
+    public function __construct(string $imageUrl, array $crop = [])
+    {
+        $this->imageUrl = $imageUrl;
+        $this->crop = $crop;
+    }
+
+    public function geturl(): string
```
Daimona: I think this should be getUrl or getURL

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
+<?php
```
Daimona, line 1: Maybe this should go to its own namespace, e.g. Image?

src/Engine/EngineBase.php
```diff
@@ -201,4 +214,41 @@ protected function getInvalidLangsWarning(array $invalidLangs): string
...
     }
+
+    /**
+     * @param string $imageUrl The original image URL.
+     * @param int[] $crop Array with keys `x, `y`, `width` and `height`.
+     * @param ?bool $fetch Whether to download the image or not.
+     * @return Image
+     * @throws OcrException If the image couldn't be fetched.
+     */
+    public function getImage(string $imageUrl, array $crop, ?bool $fetch = false): Image
```
Daimona: Suggest using class constants for $fetch, e.g. FETCH_DOWNLOAD or something (naming things is hard...)

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+class Image
+{
+
+    /** @var string Original URL of image. */
+    private $imageUrl;
+
+    /** @var int[] Array of floats with keys: `x`, `y`, `width`, and `height`. */
+    private $crop;
+
+    /** @var string Image data. */
```
Daimona: Should be string|null, as this isn't set in the constructor

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function getCrop(): Crop
+    {
+        return new Crop(
+            new Point($this->crop['x'], $this->crop['y']),
+            new Box($this->crop['width'], $this->crop['height'])
+        );
+    }
+
+    public function getData(): string
```
Daimona, line 59: In theory, we're not guaranteed to respect the signature requirements, if getData() is called without setData() being called first. I see two main options:

1 - (preferred, but unsure if it would work) have a single method that returns the "identifier" of an image. This could be an URL or image data, and the caller wouldn't know. That would avoid the needsCropping() checks, but assumes that every caller can handle the data and URL cases transparently, whereas tesseract only accepts image data.

2 - (easier) Document that setData() is required before getData(), and throw an exception from getData() if $this->data is null. There could also be an hasData() method that callers should use beforehand, instead of needsCropping().

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+    /** @var string Original URL of image. */
+    private $imageUrl;
+
+    /** @var int[] Array of floats with keys: `x`, `y`, `width`, and `height`. */
+    private $crop;
+
+    /** @var string Image data. */
+    private $data;
+
+    /** @var int Image size in bytes. */
```
Daimona: int|null

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->data = $data;
+    }
+
+    /**
+     * Get the image data size in bytes.
+     * @return int
+     */
+    public function getSize(): int
+    {
+        return $this->size ?? strlen($this->data);
```
Daimona, line 81: This is also implicitly assuming that $this->data is set if $this->size is not. I think we could either enforce calling getSize() (by throwing if $this->size is null), or changing this to

```
if ( $this->size !== null ) {
    return $this->size;
}
if ( $this->data !== null ) {
    return strlen( $this->data );
}

throw ...
```

src/Engine/Image.php
```diff
@@ -0,0 +1,76 @@
...
+
+    public function geturl(): string
+    {
+        return $this->imageUrl;
+    }
+
+    public function needsCropping(): bool
+    {
+        return isset($this->crop['width']) && $this->crop['width'] > 0
+            && isset($this->crop['height']) && $this->crop['height'] > 0;
```
Daimona, line 43: Should this be an OR? I.e.

```
return ( isset($this->crop['width']) && $this->crop['width'] > 0 )
            || ( isset($this->crop['height']) && $this->crop['height'] > 0 );
```

This would only make a difference if width or height is 0, in which case we're currently using the whole image. It shouldn't matter for ordinary usage anyway.

Daimona (commented):
Looking good overall, already tested a bit. A few comments, mostly nits.

src/Engine/Image.php
samwilson: Done. It was a typo. :-)

src/Engine/EngineBase.php
samwilson: Done. Went with DO_DOWNLOAD_IMAGE because it seems to read better. Doesn't really need to be a bool any more, but it doesn't not need to be either.

src/Engine/Image.php
samwilson: It'd be the only thing in that NS then. It feels okay here, to me: it's a value class that's only used by the Engines. Very happy to move if you feel strongly that it'd be better though!

src/Engine/Image.php
samwilson: Done.

src/Engine/Image.php
samwilson: Done.

Not sure why phan doesn't pick these up.

src/Engine/Image.php
samwilson: I wondered about that, but if you supply a `0` height or width for cropping you get an error, e.g.:

> Length of either side cannot be 0 or negative, current size is 140x0

It does mean that if someone only supplies one dimension they get the whole image back, but I think that's okay. The nicer way might be to default the other dimension to the full extent of the image, but to do that we have to download the image and this method is used to determine whether or not we want to download it, so it all gets a bit more complicated — for a use case that feels pretty rare.

src/Engine/Image.php
samwilson: Good point. I've gone with the latter, because as you say although it'd be nice for the Engine to not have to know about URL vs data, it doesn't seem very possible.

src/Engine/Image.php
samwilson: Actually, it should probably throw if data is null, because an Image shouldn't ever have a size but no data; the size is the data-size after all. Does that sound correct?

samwilson (commented):
I've also added some reset stuff, for when the URL is changed in the form (we don't want to use the same crop coords then, and it also doesn't make sense to be able to change the URL *and* select a crop region in the old image).

src/Engine/Image.php
Daimona: By design, phan is quite lenient with union types. For instance, if X can have type S|T and doing some operation requires type S, phan will not complain about X possibly being a T (at least, that's how it used to work); the idea behind that is to have a low false positive rate while still being helpful. So I guess it doesn't even have a check for missing null in uninitialized class members, as it may increase the FP rate.

src/Engine/Image.php
Daimona: Yeah, especially if we agree that in the URL case, size means "size of the data pointed by the URL".

src/Engine/Image.php
Daimona: Oh, I didn't see that error. But yeah, it doesn't feel necessary to complicate this. Thus, it's fine to leave the code as it currently is.

src/Engine/Image.php
Daimona: I think it'd be especially useful if we plan to add more classes to that NS, otherwise, as you said, I don't think it's necessary. One example might be a new Exception class used in Image to assert preconditions on `$this->data`. But again, it's just a minor change and I agree with you that we can leave it here.

src/Engine/Image.php
```diff
@@ -0,0 +1,88 @@
...
+
+    public function hasData(): bool
+    {
+        return null !== $this->data;
+    }
+
+    public function getData(): string
+    {
+        if (null === $this->data) {
+            throw new Exception('Image::setData() must be called before getData()');
```
Daimona: Suggest LogicException or RuntimeException for this and line 79

Daimona (commented):
New changes look good, haven't tested yet.

src/Engine/Image.php
samwilson: Done.

Daimona:
Tested, works as expected. I originally thought that the image-resetting thing didn't work, because nothing happened when I changed the URL. Then I realized that the change event is only fired when the input loses focus. I think the reset itself is fine, since if you change the URL, you're most certainly going to transcribe another image. If any issue comes up that we didn't expect, the UI can be changed later on.

PR #57, opened 2021-08-23
Add drag-mode switching buttons

samwilson:
Add radio buttons to toggle between 'crop' and 'move' drag modes,
and a separate 'Transcribe area' button for submitting to crop
the image.

The change to the copy-button is because it's now inside the form.

Bug: https://phabricator.wikimedia.org/T288672


assets/styles/app.css
```diff
@@ -69,8 +69,7 @@ fieldset,
 	margin-top: 40px;
 }
 
-.copy-button {
-	position: absolute;
-	right: 30px;
-	top: 15px;
+.buttons {
```
Daimona: I think 'buttons' might be a bit generic. I tried to come up with a better name, but I don't have good ideas at the moment. Perhaps 'output-buttons'?

templates/output.html.twig
```diff
@@ -39,28 +39,48 @@
...
+            <div class="ocr-output">
+                <hr/>
+                <p>{{ msg('drag-help') }}</p>
+                <div class="row">
+                    <div class="col-md-6">
+                        <div class="buttons">
+                            <div class="btn-group" role="group">
+                                <button type="button" class="btn btn-default active drag-mode drag-mode-crop"
+                                        title="{{ msg('drag-mode-crop') }}" data-drag-mode="crop" disabled>
+                                    <img src="{{ asset('build/images/Crop_-_The_Noun_Project.svg') }}" height="20" width="20" />
```
Daimona: Should this have an `alt` attribute (e.g. `alt="Crop"`) for cross-device compat? Unsure how that would work inside a button, though.

Daimona (commented):
As a general comment, I note that the new buttons don't work (obviously) if JavaScript is disabled. What do you think about having a `nojs` class in the html element, and remove it on page ready (like for MW)? Then add

```
nojs .buttons {
    display: none;
}
```

assets/styles/app.css
samwilson: Done.

templates/output.html.twig
samwilson: Done. I think it's fine inside a button.

samwilson (commented):
> As a general comment, I note that the new buttons don't work (obviously) if JavaScript is disabled. What do you think about having a `nojs` class in the html element, and remove it on page ready (like for MW)? Then add
> 
> ```
> nojs .buttons {
>     display: none;
> }
> ```

Great idea, done. Also applied it to the copy-to-clipboard button.

samwilson:
I've also updated to @nayoub 's suggestion of starting in move mode, and requiring a click on crop to start cropping. This means that the mode buttons don't start as disabled.

assets/app.js
```diff
@@ -73,9 +77,13 @@ $(function () {
         x = document.querySelector('[name="crop[x]"]'),
```
Daimona: I didn't notice earlier, but now these elements don't exist when there's no output, so there's a JS error when you first load the page (`Cannot read property 'value' of null` at line 80). I guess the whole cropping thing should be inside a conditional

Daimona (commented):
New changes look good, just one JS issue (and a merge conflict to resolve)

assets/app.js
samwilson: Good point, done. Sorry I didn't notice that earlier!

Daimona:
Thank you, everything looks good. I was thinking for a future change the "transcribe area" button could do its thing via AJAX and show the result without reloading the page, but it's just a minor thing that we don't have to care about for the time being.

PR #58, opened 2021-08-25
Move API docs to Swagger

samwilson:
Update the README docs, and move those about the API into method
annotations in the OcrController so they can be exposed via
auto-generated API docs with NelmioApiDocBundle.

Bug: https://phabricator.wikimedia.org/T285513


src/Controller/OcrController.php
```diff
@@ -177,8 +178,40 @@ public function homeAction(): Response
...
+     *     name="image",
+     *     in="query",
+     *     description="The image URL.",
+     *     @OA\Schema(type="string")
+     * )
+     * @OA\Parameter(
+     *     name="langs",
+     *     in="query",
+     *     description="List of language codes.
+     *         Can be left empty, in which case the engine will do its best (useful for unsupported languages).",
```
Daimona: This asterisk is appearing verbatim in the documentation. I think it doesn't support line breaks.

src/Controller/OcrController.php
```diff
@@ -177,8 +178,40 @@ public function homeAction(): Response
...
 
     /**
-     * @Route("/api", name="api")
-     * @Route("/api.php", name="apiPhp")
+     * @Route("/api", name="api", methods={"GET", "POST"})
+     * @Route("/api.php", name="apiPhp", methods={"GET", "POST"})
+     * @OA\Parameter(
+     *     name="engine",
+     *     in="query",
+     *     description="The engine to use, either `tesseract` or `google`.",
```
Daimona, line 186: The interactive input field for this is using "$name - $description" as placeholder, which doesn't fit into the field. Would it be possible to set it to some other value, e.g. "tesseract"? Same for the other two fields.

src/Controller/OcrController.php
```diff
@@ -177,8 +178,40 @@ public function homeAction(): Response
...
+     *         Can be left empty, in which case the engine will do its best (useful for unsupported languages).",
+     *     @OA\JsonContent(type="array", @OA\Items(type="string"))
+     * )
+     * @OA\Parameter(
+     *     name="psm",
+     *     in="query",
+     *     description="The Page Segmentation Mode for Tesseract.",
+     *     @OA\Schema(type="int")
+     * )
+     * @OA\Parameter(
```
Daimona, line 210: Noting that this one doesn't have an interactive input at all, but I guess it's not important.

src/Controller/OcrController.php
```diff
@@ -196,7 +229,13 @@ public function apiAction(): JsonResponse
...
 
     /**
-     * @Route("/api/available_langs", name="apiLangs")
+     * @Route("/api/available_langs", name="apiLangs", methods={"GET"})
+     * @OA\Parameter(
+     *     name="engine",
+     *     in="query",
+     *     description="The engine to use, either `tesseract` or `google`.",
+     *     @OA\Schema(type="string")
+     * )
```
Daimona, line 241: Missing Response documentation, could be e.g.

```
@OA\Response(response=200, description="List of available langs in JSON format.")
```

src/Controller/OcrController.php
Daimona: Found https://github.com/zircote/swagger-php/issues/326

Daimona:
(As a reminder, POST should be removed from /api as we already discussed)

src/Controller/OcrController.php
samwilson: I've added an example, which is then used as the placeholder.

src/Controller/OcrController.php
samwilson: Oops! Fixed.

src/Controller/OcrController.php
samwilson: Done.

src/Controller/OcrController.php
samwilson: You have to click 'add item' to get an input for this, but actually it still doesn't work because there's no way to add the keys. I'm not sure it matters too much, and I think to fix it we need a value object that can be used as a Model (which feels like a bigger change than this; a good thing to keep in mind next time we're building a tool's API though).

samwilson:
> (As a reminder, POST should be removed from /api as we already discussed)

Done.

I also moved the API docs endpoint itself to be at `/openapi.json` as recommended by the spec. I guess it's more discoverable there by automated tools. I tried to see if there was some standard of linking to it from the HTML of the site itself (like RSD does), but couldn't find any info.

src/Controller/OcrController.php
Daimona: Ah, I see. Totally not important.

Daimona (approved):
Nice, thank you!

PR #59, opened 2022-12-08
Allow protocol relative URLs for the image

samwilson:
Both http and https are already allowed, and the actual validation is done based on hostname so there shouldn't be any issue with making this regex more lenient.

Bug: T324740


samwilson:
I'm fixing the Phan failure here by upgrading Phan, see #60.

samwilson:
Oh well, the Phan upgrade didn't fix it; have suppressed it for now. Will follow up later because we need to get this merged.

PR #63, opened 2023-01-31
Create OAuth CLI command for transkribus auth

Parthiv-M:
Created app:transkribus command that takes username and password and returns an access token to use
the transkribus API

Bug: T327617

### DIFF
--- a/.env
+++ b/.env
@@ -33,4 +33,3 @@ APP_LOG_SUBJECT="[Wikimedia OCR]"
 APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
-
--- a/.env
+++ b/.env
@@ -33,3 +33,5 @@ APP_LOG_SUBJECT="[Wikimedia OCR]"
 APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
+
+CLIENT_ID=
--- a/.env
+++ b/.env
@@ -34,4 +34,4 @@ APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
 
-CLIENT_ID=
+APP_TRANSKRIBUS_CLIENT_ID=
--- a/.env
+++ b/.env
@@ -34,4 +34,3 @@ APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
 
-APP_TRANSKRIBUS_CLIENT_ID=
--- a/composer.json
+++ b/composer.json
@@ -17,6 +17,7 @@
         "symfony/console": "5.2.*",
         "symfony/dotenv": "5.2.*",
         "symfony/framework-bundle": "^5.4",
+        "symfony/http-client": "^5.4",
         "symfony/mailer": "^5.2",
         "symfony/monolog-bundle": "^3.7",
         "symfony/property-info": "5.2.*",
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,6 @@
         "symfony/console": "5.2.*",
         "symfony/dotenv": "5.2.*",
         "symfony/framework-bundle": "^5.4",
-        "symfony/http-client": "^5.4",
         "symfony/mailer": "^5.2",
         "symfony/monolog-bundle": "^3.7",
         "symfony/property-info": "5.2.*",
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -51,6 +51,10 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.exception }
 
+    App\Command\TranskribusAuthCommand:
+        tags:
+            - { name: 'console.command', command: 'app:transkribus' }
+
     # Vendor services for autowiring
     thiagoalessio\TesseractOCR\TesseractOCR:
 
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -52,8 +52,8 @@ services:
             - { name: kernel.event_listener, event: kernel.exception }
 
     App\Command\TranskribusAuthCommand:
-        tags:
-            - { name: 'console.command', command: 'app:transkribus' }
+        arguments:
+            $clientId: '%env(APP_TRANSKRIBUS_CLIENT_ID)%'
 
     # Vendor services for autowiring
     thiagoalessio\TesseractOCR\TesseractOCR:
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -51,10 +51,6 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.exception }
 
-    App\Command\TranskribusAuthCommand:
-        arguments:
-            $clientId: '%env(APP_TRANSKRIBUS_CLIENT_ID)%'
-
     # Vendor services for autowiring
     thiagoalessio\TesseractOCR\TesseractOCR:
 
--- a/src/Command/TranskribusAuthCommand.php
+++ b/src/Command/TranskribusAuthCommand.php
@@ -15,9 +15,6 @@ class TranskribusAuthCommand extends Command
     /** @var HttpClientInterface The interface to make http requests */
     private $client;
 
-    /** @var string The client ID passed to the transkribus API */
-    private $clientId;
-
     public function __construct(HttpClientInterface $client)
     {
         parent::__construct();
--- a/src/Command/TranskribusAuthCommand.php
+++ b/src/Command/TranskribusAuthCommand.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace App\Command;
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputDefinition;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class TranskribusAuthCommand extends Command
+{
+
+    private $client;
+
+    public function __construct(HttpClientInterface $client)
+    {
+        parent::__construct();
+        $this->client = $client;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setHelp('This command allows you to retrieve and store an access token for testing the Trankribus engine');
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $helper = $this->getHelper('question');
+        $io = new SymfonyStyle($input, $output);
+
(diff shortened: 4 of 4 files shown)


.env
```diff
@@ -33,3 +33,5 @@ APP_LOG_SUBJECT="[Wikimedia OCR]"
 APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
+
+CLIENT_ID=
```
samwilson: Should be e.g. `APP_TRANSKRIBUS_CLIENT_ID` to make it clearer what it's for and to use the standard `APP` prefix.

composer.json
```diff
@@ -17,6 +17,7 @@
         "symfony/console": "5.2.*",
         "symfony/dotenv": "5.2.*",
         "symfony/framework-bundle": "^5.4",
+        "symfony/http-client": "^5.4",
```
samwilson: This is already a requirement of `wikimedia/toolforge-bundle` so no need to add it separately.

config/services.yaml
```diff
@@ -51,6 +51,10 @@ services:
         tags:
             - { name: kernel.event_listener, event: kernel.exception }
 
+    App\Command\TranskribusAuthCommand:
+        tags:
+            - { name: 'console.command', command: 'app:transkribus' }
```
samwilson: The command name could be defined in `TranskribusAuthCommand::configure()` instead of here (with `$this->setName( 'app:transkribus' )`). That keeps this file free of unnecessary things, and puts the definition closer to where it's used.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,81 @@
...
+    public function __construct(HttpClientInterface $client)
+    {
+        parent::__construct();
+        $this->client = $client;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setHelp('This command allows you to retrieve and store an access token for testing the Trankribus engine');
```
samwilson: It's not just for testing. Suggest: "…an access token for the Transkribus API."

Also, this should use `setDescription()` instead of `setHelp()`, so this is displayed in the command list and also at the top of the help output. Help is for longer information and appears at the end of the help output.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,81 @@
...
+        $helper = $this->getHelper('question');
+        $io = new SymfonyStyle($input, $output);
+
+        $userNameQuestion = new Question('Enter username: ');
+        $userName = $helper->ask($input, $output, $userNameQuestion);
+
+        $passwordQuestion = new Question('Enter password: ');
+        $password = $helper->ask($input, $output, $passwordQuestion);
+
+        $clientId = $_ENV['CLIENT_ID'];
```
samwilson: Client ID should be injected in the constructor, with something like the following in `services.yaml`:

```yaml
        arguments:
            $clientID: '%env(APP_TRANSKRIBUS_CLIENT_ID)%'
```

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,81 @@
...
+                        ]
+                    );    
+        $statusCode = $token_response->getStatusCode();
+        if ( $statusCode == 200 ) {
+            $content = json_decode($token_response->getContent());
+            $access_token = $content->{'access_token'};
+            $output->writeln([
+                'Your access token is: '.$access_token,
+                $io->newline(),
+                '--- Please copy and store the access token for future use ---'
```
samwilson: This should explain how the token needs to be added to `.env.local`.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,81 @@
...
+                                'password' => $password,
+                                'client_id' => $clientId,
+                                'scope' => 'offline_access'
+                            ]
+                        ]
+                    );    
+        $statusCode = $token_response->getStatusCode();
+        if ( $statusCode == 200 ) {
+            $content = json_decode($token_response->getContent());
+            $access_token = $content->{'access_token'};
```
samwilson: `$content` could at this point be null; this should be checked for.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,81 @@
...
+            $output->writeln([
+                'Your access token is: '.$access_token,
```
samwilson: How long does this access token live for? My understanding was that this is an OAuth flow, and so the initial request results in a request token which then needs to be exchanged for an access token. That's not the case?

samwilson (changes requested):
Looking good! I've left a few comments.

Don't forget to run `composer test` before pushing.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+        $helper = $this->getHelper('question');
+        $io = new SymfonyStyle($input, $output);
+
+        $userNameQuestion = new Question('Enter username: ');
+        $userName = $helper->ask($input, $output, $userNameQuestion);
+
+        $passwordQuestion = new Question('Enter password: ');
+        $password = $helper->ask($input, $output, $passwordQuestion);
+
+        $token_response = $this->client
```
samwilson: Although it doesn't seem to be enforced by phpcs, variables should be camelCase. I've created T328730 for this, but you should fix these here.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $helper = $this->getHelper('question');
+        $io = new SymfonyStyle($input, $output);
+
+        $userNameQuestion = new Question('Enter username: ');
+        $userName = $helper->ask($input, $output, $userNameQuestion);
+
+        $passwordQuestion = new Question('Enter password: ');
```
samwilson, line 39: $passwordQuestion should be a hidden question, so that the password isn't shown when entered.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+                                'client_id' => $this->clientId,
+                                'scope' => 'offline_access',
+                            ],
+                        ]
+                    );
+        $statusCode = $token_response->getStatusCode();
+        if (200 === $statusCode) {
+            $content = json_decode($token_response->getContent());
+            if (is_null($content)) {
+                $output->writeln('Transkribus API returned null');
```
samwilson: It didn't necessarily return `null`: it returned something that didn't decode as json.

Also, this should be using `$io->error()`.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+                return Command::FAILURE;
+            } else {
+                $access_token = $content->{'access_token'};
+                $output->writeln([
```
samwilson: This should be `$io->writeln()`, and it's probably neater to just call it twice instead of using `$io->newline()` below.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+                    'Your access token is: '.$access_token,
+                    $io->newline(),
+                    '--- Please copy and add it to your .env.local as APP_TRANSKRIBUS_ACCESS_TOKEN ---',
+                ]);
+                return Command::SUCCESS;
+            }
+        } elseif (401 === $statusCode) {
+            $output->writeln('Error Code '.$statusCode.' :: Credentials are incorrect!');
```
samwilson: This and the error below should both use `$io->error()`.

.env
```diff
@@ -33,3 +33,5 @@ APP_LOG_SUBJECT="[Wikimedia OCR]"
 APP_IMAGE_HOSTS=upload.wikimedia.org,upload.wikimedia.beta.wmflabs.org
 
 APP_CACHE_TTL=3600 # 1 hour
+
+APP_TRANSKRIBUS_CLIENT_ID=
```
samwilson: Sorry I didn't ask this earlier, but why is this configurable? It appears to only have a single possible value, of `processing-api-client`. Is that right? Can this just be hardcoded?

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,86 @@
...
+            ->setName('app:transkribus')
+            ->setDescription('This command allows you to retrieve and store an access token for the Trankribus API');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $helper = $this->getHelper('question');
+        $io = new SymfonyStyle($input, $output);
+
+        $userNameQuestion = new Question('Enter username: ');
```
samwilson, line 36: Before asking for username and continuing, this should look for `APP_TRANSKRIBUS_ACCESS_TOKEN` and if it's found it should use that to check and report the current authentication status.

samwilson (changes requested):

src/Command/TranskribusAuthCommand.php
Parthiv-M: I'm not sure about how to check the current authentication status using `APP_TRANSKRIBUS_ACCESS_TOKEN`. I'll take a look at it and try including that in the next commit

Parthiv-M:
Oh, I suppose the unwanted whitespace in the `.env` file can be removed as well

src/Command/TranskribusAuthCommand.php
samwilson: Sure that makes sense.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,118 @@
...
+                        $io->writeln('--- Please copy and add it to your .env.local'.
+                                                'as APP_TRANSKRIBUS_ACCESS_TOKEN ---');
+                        return Command::SUCCESS;
+                    }
+                } elseif (400 === $statusCode) {
+                    $io->error('Error Code '.$statusCode.' :: Refresh token is incorrect');
+                    return Command::FAILURE;
+                } else {
+                    $io->error('Error Code '.$statusCode.' :: Error generating access'.
+                                                'token from refresh token, try again!');
```
samwilson: There's no space between 'access' and 'token'.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,118 @@
...
+                    if (is_null($content)) {
+                        $io->error('Response from Transkribus API could not be parsed');
+                        return Command::FAILURE;
+                    } else {
+                        $accessToken = $content->{'access_token'};
+                        $io->writeln('');
+                        $io->writeln('Your access token is: '.$accessToken);
+                        $io->writeln('');
+                        $io->writeln('--- Please copy and add it to your .env.local'.
+                                                'as APP_TRANSKRIBUS_ACCESS_TOKEN ---');
```
samwilson: There's no space between '.env.local' and 'as'.

samwilson (commented):
Looking great. Couple of tiny comments but I think it's pretty much good to go.

src/Command/TranskribusAuthCommand.php
```diff
@@ -0,0 +1,118 @@
...
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TranskribusAuthCommand extends Command
+{
+
+    /** @var HttpClientInterface The interface to make http requests */
+    private $client;
+
+    /** @var string The client ID passed to the transkribus API */
+    private $clientId;
```
samwilson: This is not used.

Parthiv-M: Oops, skipped that

kolakachi (approved):

PR #64, opened 2023-02-06
Remove libopenjp2-7 installation from CI

Parthiv-M:
Change the github workflow file to not install the `libopenjp2-7` package during CI check


.github/workflows/ci.yml
```diff
@@ -29,7 +29,6 @@ jobs:
       # Note: libopenjp2-7 is already installed, but it's apparently corrupted
```
samwilson: This comment is no longer valid.

samwilson:
Thanks!

PR #65, opened 2023-02-06
Add Transkribus engine to the OCR tool

kolakachi:
Added Transkribus to the OCR Tool's API
as an option for the engine.

Bug: T326389

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -44,9 +44,10 @@ services:
     
     App\Engine\TranskribusEngine:
         parent: App\Engine\EngineBase
+            
+    App\Engine\TranskribusClient:
         arguments:
-            $username: '%env(USERNAME)%'
-            $password: '%env(PASSWORD)%'
+            $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
 
     App\EventListener\ExceptionListener:
         arguments:
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -41,10 +41,10 @@ services:
         parent: App\Engine\EngineBase
         arguments:
             $keyFile: '%env(APP_GOOGLE_KEYFILE)%'
-    
+
     App\Engine\TranskribusEngine:
         parent: App\Engine\EngineBase
-            
+
     App\Engine\TranskribusClient:
         arguments:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -41,6 +41,12 @@ services:
         parent: App\Engine\EngineBase
         arguments:
             $keyFile: '%env(APP_GOOGLE_KEYFILE)%'
+    
+    App\Engine\TranskribusEngine:
+        parent: App\Engine\EngineBase
+        arguments:
+            $username: '%env(USERNAME)%'
+            $password: '%env(PASSWORD)%'
 
     App\EventListener\ExceptionListener:
         arguments:
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -57,5 +57,10 @@
 	"tesseract-psm-12": "Form option for Tesseract page segmentation mode.",
 	"tesseract-psm-13": "Form option for Tesseract page segmentation mode.",
 	"tesseract-param-error": "Error message displayed when invalid values for Tesseract options are submitted.\n\nParameters:\n* $1 – the form label for the option. Currently, the only possible value here is {{msg-wm|wikimedia-ocr-tesseract-psm-label}}.\n* $2 – The value that was given.\n* $3 – the maximum value for the option (this will be an integer).",
-	"tesseract-internal-error": "Generic error message displayed when the tesseract command fails."
+	"tesseract-internal-error": "Generic error message displayed when the tesseract command fails.",
+	"transkribus-error-401": "Error Code '$1' :: The request is not authorized",
+    "transkribus-error-default": "Error Code '$1' ::  Error initiating upload process, try again!",
+    "transkribus-error-empty-response": "Could not parse result from Transkribus API",
+	"transkribus-error-init-process-failed": "Failed to init Transkribus process",
+	"transkribus-error-failed-process-status": "Failed to retrieve Transkribus process status"
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -18,6 +18,7 @@
 	"engine": "Form label for the 'engine' field.",
 	"engine-name-google": "Option text for the 'google' engine.",
 	"engine-name-tesseract": "Option text for the 'tesseract' engine.",
+	"engine-name-transkribus": "Option text for the 'transkribus' engine.",
 	"engine-not-found-warning": "Error message displayed when an invalid engine is requested.\n\nParameter:\n* $1 – the requested engine.\n* $2 – the default engine that is being used instead.",
 	"engine-invalid-langs-warning": "Warning message displayed when invalid or unsupported languages are requested.\n\nParameter:\n* $1 – a comma-separated list of invalid languages.",
 	"submit": "Submit button text for when the whole image is being submitted.",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -58,9 +58,9 @@
 	"tesseract-psm-13": "Form option for Tesseract page segmentation mode.",
 	"tesseract-param-error": "Error message displayed when invalid values for Tesseract options are submitted.\n\nParameters:\n* $1 – the form label for the option. Currently, the only possible value here is {{msg-wm|wikimedia-ocr-tesseract-psm-label}}.\n* $2 – The value that was given.\n* $3 – the maximum value for the option (this will be an integer).",
 	"tesseract-internal-error": "Generic error message displayed when the tesseract command fails.",
-	"transkribus-error-401": "Error Code '$1' :: The request is not authorized",
-    "transkribus-error-default": "Error Code '$1' ::  Error initiating upload process, try again!",
-    "transkribus-error-empty-response": "Could not parse result from Transkribus API",
-	"transkribus-error-init-process-failed": "Failed to init Transkribus process",
-	"transkribus-error-failed-process-status": "Failed to retrieve Transkribus process status"
+	"transkribus-unauthorized-error": "Error message displayed when Transkribus access token has expired or login credentials are invalid.",
+    "transkribus-default-error": "Generic error message displayed when Transkribus API returns error status different from 401.",
+    "transkribus-empty-response-error": "Error message displayed when Transkribus API returns an empty response body.",
+	"transkribus-init-process-error": "Error message displayed when Transkribus API fails to initialize upload process.",
+	"transkribus-failed-process-error": "Error message displayed when Transkribus API process status returns FAILED."
 }
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -8,6 +8,7 @@ use App\Engine\EngineFactory;
 use App\Engine\EngineResult;
 use App\Engine\GoogleCloudVisionEngine;
 use App\Engine\TesseractEngine;
+use App\Engine\TranskribusEngine;
 use App\Exception\EngineNotFoundException;
 use Krinkle\Intuition\Intuition;
 use OpenApi\Annotations as OA;
@@ -27,7 +28,7 @@ class OcrController extends AbstractController
     /** @var Intuition */
     protected $intuition;
 
-    /** @var TesseractEngine|GoogleCloudVisionEngine */
+    /** @var TesseractEngine|GoogleCloudVisionEngine|TranskribusEngine */
     protected $engine;
 
     /** @var string The default OCR engine. */
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -185,7 +185,7 @@ class OcrController extends AbstractController
      * @OA\Parameter(
      *     name="engine",
      *     in="query",
-     *     description="The engine to use, either `tesseract` or `google`.",
+     *     description="The engine to use, either `tesseract` or `google` or transkribus.",
      *     example="tesseract",
      *     @OA\Schema(type="string")
      * )
@@ -237,7 +237,7 @@ class OcrController extends AbstractController
      * @OA\Parameter(
      *     name="engine",
      *     in="query",
-     *     description="The engine to use, either `tesseract` or `google`.",
+     *     description="The engine to use, either `tesseract` or `google` or transkribus.",
      *     example="tesseract",
      *     @OA\Schema(type="string")
      * )
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -185,7 +185,7 @@ class OcrController extends AbstractController
      * @OA\Parameter(
      *     name="engine",
      *     in="query",
-     *     description="The engine to use, either `tesseract` or `google` or transkribus.",
+     *     description="The engine to use, either `tesseract` or `google` or `transkribus`.",
      *     example="tesseract",
      *     @OA\Schema(type="string")
      * )
@@ -237,7 +237,7 @@ class OcrController extends AbstractController
      * @OA\Parameter(
      *     name="engine",
      *     in="query",
-     *     description="The engine to use, either `tesseract` or `google` or transkribus.",
+     *     description="The engine to use, either `tesseract` or `google` or `transkribus`.",
      *     example="tesseract",
      *     @OA\Schema(type="string")
      * )
--- a/src/Engine/TranskribusClient.php
+++ b/src/Engine/TranskribusClient.php
@@ -59,7 +59,7 @@ class TranskribusClient
             return $processId;
         }
 
-        throw new OcrException('transkribus-error-init-process-failed');
+        throw new OcrException('transkribus-init-process-error');
     }
 
     /**
@@ -74,7 +74,7 @@ class TranskribusClient
         $textResult = '';
 
         if ('FAILED' === $content->status) {
-            throw new OcrException('transkribus-error-failed-process-status');
+            throw new OcrException('transkribus-failed-process-error');
         }
 
         if ('FINISHED' === $content->status) {
@@ -93,16 +93,16 @@ class TranskribusClient
         switch ($statusCode) {
             case 0:
                 throw new OcrException(
-                    'transkribus-error-empty-response'
+                    'transkribus-empty-response-error'
                 );
             case 401:
                 throw new OcrException(
-                    'transkribus-error-401',
+                    'transkribus-unauthorized-error',
                     [$statusCode]
                 );
             default:
                 throw new OcrException(
-                    'transkribus-error-default',
+                    'transkribus-default-error',
                     [$statusCode]
                 );
         }
--- a/src/Engine/TranskribusClient.php
+++ b/src/Engine/TranskribusClient.php
@@ -60,40 +60,22 @@ class TranskribusClient
                 ],
             ];
         }
-        $response = $this->httpClient
-        ->request(
-            'POST',
-            self::PROCESSES_URL,
-            [
-                'headers' => [
-                    'Content-Type' => 'application/json',
-                    'Authorization' => 'Bearer '.$this->accessToken,
-                ],
-                'json' => [
-                    'config' => $this->config,
-                    'image' => [
-                        'imageUrl' => $imageURL,
-                    ],
-                ],
-            ]
-        );
 
-        $statusCode = $response->getStatusCode();
-        if (200 === $statusCode) {
-            $content = json_decode($response->getContent());
-            
-            if (is_null($content)) {
-                $this->setErrorMessage(0);
-                return $this;
-            }
-            
+        $jsonBody = [
+            'config' => $this->config,
+            'image' => [
+                'imageUrl' => $imageURL,
+            ],
+        ];
+
+        $content = $this->request('POST', self::PROCESSES_URL, $jsonBody);
+        $parsedContent = (array) $content;
+        if (!empty($parsedContent)) {
             if ($content->{'status'} === 'CREATED') {
                 $this->processId = $content->{'processId'};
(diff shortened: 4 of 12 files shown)


src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,277 @@
...
+                                'client_id' => 'processing-api-client',
+                                'scope' => 'offline_access',
+                            ],
+                        ]
+                    );
+        $statusCode = $response->getStatusCode();
+        if (200 === $statusCode) {
+            $content = json_decode($response->getContent());
+            if (is_null($content)) {
+                $message = 'Transkribus API returned null';
```
Parthiv-M, line 86: I actually changed this error message later on, can we update this here as well? 😅

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,277 @@
...
+                    ]
+                ],
+            ]
+        );
+
+        $statusCode = $response->getStatusCode();
+        if (200 === $statusCode) {
+            $content = json_decode($response->getContent());
+            if (is_null($content)) {
+                $message = 'Transkribus API returned null';
```
Parthiv-M, line 131: Same thing about the error message here as well

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,277 @@
...
+                        'refresh_token' => $refreshToken,
+                    ],
+                ]
+            );
+            $statusCode = $response->getStatusCode();
+            if (200 === $statusCode) {
+                $content = json_decode($response->getContent());
+
+                if (is_null($content)) {
+                    $message = 'Transkribus API returned null';
```
Parthiv-M, line 237: - Need to modify the error message again
- I think this is missing a call to `setErrorMessage()`

Parthiv-M (dismissed):
Need to change the auth flow once the CLI command is complete

config/services.yaml
```diff
@@ -41,6 +41,12 @@ services:
         parent: App\Engine\EngineBase
         arguments:
             $keyFile: '%env(APP_GOOGLE_KEYFILE)%'
+    
+    App\Engine\TranskribusEngine:
+        parent: App\Engine\EngineBase
+        arguments:
+            $username: '%env(USERNAME)%'
```
samwilson: This should only be reading in the `APP_TRANSKRIBUS_ACCESS_TOKEN` from `.env` and not the username and password. PR #63 needs to be merged first, of course.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,86 @@
...
+        Intuition $intuition,
+        string $projectDir,
+        HttpClientInterface $httpClient
+    ) {
+        parent::__construct($intuition, $projectDir, $httpClient);
+        $credentials = [
+            'username' => $username,
+            'password' => $password
+        ];
+        $this->transkribusClient = new TranskribusClient($httpClient, $credentials);
```
samwilson: The TranskribusClient object shouldn't be created here, it should be injected as a constructor parameter. That way, the service container handles instantiation.

templates/output.html.twig
```diff
@@ -23,6 +23,12 @@
                         {{ msg('engine-name-tesseract') }}
                     </label>
                 </div>
+                <div class="radio">
+                    <label>
+                        <input type="radio" name="engine" value="transkribus" {% if engine == 'transkribus' %}checked{% endif %} />
+                       Transkribus OCR
```
samwilson: Needs internationalization (i.e. using `msg()` and adding to `en.json` and `qqq.json`). Copy how the others are handled.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->checkImageUrl($imageUrl);
+
+        [ $validLangs, $invalidLangs ] = $this->filterValidLangs($langs, $invalidLangsMode);
+
+        $image = $this->getImage($imageUrl, $crop);
+        $imageUrl = $image->getUrl();
+        $response = $this->transkribusClient->initProcess($imageUrl, []);
+
+        if ($response->getHasError()) {
+            file_put_contents('initp.json', "init process failed");
```
samwilson: What is this file for? It doesn't seem to be used anywhere.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,277 @@
...
+
+    public function refereshAccessToken(){
+        if(!file_exists('transkribus_credentials.json')){
+            $this->setAccessToken();
+            
+        }else{
+            $data = file_get_contents('transkribus_credentials.json');
+            $content = json_decode($data);
+            $refreshToken = $content->{'refresh_token'};
+            $response = $this->httpClient
```
samwilson: The pattern of making a request, looking at its response code and content, and returning errors, is repeated a few times. It's probably worth abstracting into a separate private function e.g. `TranskribusClient::request()`. That way, the TranskribusAuthCommand could also have TranskribusClient injected and use it to make the requests there.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,277 @@
...
+        if(file_exists('transkribus_credentials.json')){
+            $data = file_get_contents('transkribus_credentials.json');
+            $content = json_decode($data);
+            $this->accessToken = $content->{'access_token'};
+        }else{
+           $this->retrieveAccessToken();
+        }        
+    }
+
+    private function retrieveAccessToken(){
```
samwilson: The authentication process will be done by #63 so doesn't need to be repeated here.

samwilson (commented):
This is looking good! I've not gone through it deeply, but just made few comments about things that I noticed. The main thing now I think is to get #63 merged, and then rebase this on main so it doesn't have to do any of the OAuth stuff. Then we can look into the new engine workflow properly.

samwilson:
Oh, and the other thing: make sure you run `composer test` and `npm run test` before pushing. That'll catch some of the simpler things.

src/Engine/TranskribusEngine.php
kolakachi: Will remove that, was trying to debug

Parthiv-M (approved):

config/services.yaml
```diff
@@ -41,6 +41,13 @@ services:
         parent: App\Engine\EngineBase
         arguments:
             $keyFile: '%env(APP_GOOGLE_KEYFILE)%'
+    
+    App\Engine\TranskribusEngine:
+        parent: App\Engine\EngineBase
+
```
samwilson: Trailing spaces here and above on line 44.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
...
+    private $reponseHasError = false;
+    
+    /** @var string Transcribed result. */
+    private $textResult = '';
+    
+    /** @var $config. */
+    private $config = [];
+
+    /** @var $langs valid languages. */
+    private $langs = '';
```
samwilson: Not used.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+namespace App\Engine;
+
+use App\Exception\OcrException;
+use Krinkle\Intuition\Intuition;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TranskribusEngine extends EngineBase
+{
+    /** @var string The API key. */
+    protected $key;
```
samwilson: Not used.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+     * @param HttpClientInterface $httpClient
+     */
+    public function __construct(
+        TranskribusClient $transkribusClient,
+        Intuition $intuition,
+        string $projectDir,
+        HttpClientInterface $httpClient
+    ) {
+        parent::__construct($intuition, $projectDir, $httpClient);
+
```
samwilson: Trailing spaces.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
...
+        string $accessToken
+    ) {
+        $this->httpClient = $httpClient;
+        $this->accessToken = $accessToken;
+    }
+
+    /**
+     * @param string $imageURL
+     */
+    public function initProcess(string $imageURL): self
```
samwilson: Why is this a fluent method? (i.e. returns itself). I would suggest that this should either throw an exception in case of an error, or return the process ID (or whatever else is required).

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
...
+            if ($content->{'status'} === 'CREATED') {
+                $this->processId = $content->{'processId'};
+                $this->reponseHasError = false;
+            }
+        }
+
+        return $this;
+    }
+
+    public function retrieveProcessStatus(): self
```
samwilson: This should take a process ID as a parameter, and return the status of that process.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
...
+        $this->langs = $langs;
+    }
+
+     /**
+     * @param string $method
+     * @param string $url
+     * @param mixed[] $jsonBody
+     * @return object
+     */
+    public function request(
```
samwilson: Doesn't need to be public. Users of this class shouldn't be able to make requests directly.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
+<?php
+declare(strict_types = 1);
+
+namespace App\Engine;
+
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TranskribusClient
```
samwilson, line 9: This class shouldn't be mutable like this. At the moment, it holds response info, error messages etc. that relate to the most recently called process. This is fragile, and makes the class harder to understand from the outside. This class is responsible for being the interface to the remote API, and should make it unnecessary for people using this class to know anything about that API. It looks like it needs something like the following:

1. `initProcess(string $imageUrl): string` to get a process ID;
2. `retrieveProcess(string $processId): array` to get the process result or current status;

Could possibly also have the authentication-related stuff, but that could be a subsequent improvement:

3. `authInit($username, $password)`
4. `authRefresh($token)`

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+        array $crop,
+        ?array $langs = null
+    ): EngineResult {
+        $this->checkImageUrl($imageUrl);
+
+        [ $validLangs, $invalidLangs ] = $this->filterValidLangs($langs, $invalidLangsMode);
+
+        $image = $this->getImage($imageUrl, $crop);
+        $imageUrl = $image->getUrl();
+        $response = $this->transkribusClient->initProcess($imageUrl);
```
samwilson: This is not a response object here, it's still a TranskribusClient, so I'd suggest making the name clearer. However, I think the fluent style is not applicable here and perhaps shouldn't be used.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if ($validLangs) {
+            $response->setLanguages($this->getLangCodes($validLangs));
```
samwilson: The languages are being set here on the client object *after* the request has been sent.

The whole topic of languages/models I think could probably wait for a separate patch anyway, because it does complicate things, and you've not implemented it in this patch yet anyway.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+
+        if ($response->hasError()) {
+            throw new OcrException('transkribus-error', [$response->getErrorMessage()]);
+        }
+
+        $counter = 0;
+        while (!$response->hasError() && '' === $response->getTextResult() && 11 >$counter) {
+            $response->retrieveProcessStatus();
+            $counter++;
+            sleep(5);
```
samwilson: I think a wait of up to 55 seconds is probably too much. But we can revisit this later I'm sure.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,85 @@
...
+        if ($validLangs) {
+            $response->setLanguages($this->getLangCodes($validLangs));
+        }
+
+        if ($response->hasError()) {
+            throw new OcrException('transkribus-error', [$response->getErrorMessage()]);
+        }
+
+        $counter = 0;
+        while (!$response->hasError() && '' === $response->getTextResult() && 11 >$counter) {
```
samwilson: A result of empty string is valid and shouldn't be a loop condition. Is it not enough to look for error messages or response status?

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,179 @@
...
+
+     /**
+     * @param string[] $langs
+     */
+    public function setLanguages(array $langs): void
+    {
+        $this->langs = $langs;
+    }
+
+     /**
```
samwilson: There's an extraneous space here and on line 134.

samwilson (commented):
A few comments on various things. Mainly I think the immutability of the client class is the most important.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+        $this->httpClient = $httpClient;
+        $this->accessToken = $accessToken;
+    }
+
+    /**
+     * @param string $imageURL
+     * @param mixed[] $config
+     * @return int
+     */
+    public function initProcess(string $imageURL, array $config = []): int
```
samwilson: Is `$config` ever used? It looks like we might want to pass an `$htrId` here instead, and use that in the config (which wouldn't need to be conditionally constructed).

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+    /** @var HttpClientInterface */
+    private $httpClient;
+
+    /** @var string transkribus access token. */
+    private $accessToken = '';
+
+    /** @var $config. */
+    private $config = [];
+
+    /** @var string Transcribed result. */
```
samwilson: Wrong comment (this is the API URL, not a result).

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+class TranskribusClient
+{
+
+    /** @var HttpClientInterface */
+    private $httpClient;
+
+    /** @var string transkribus access token. */
+    private $accessToken = '';
```
samwilson: No need for a default value here, as this will always be set in the constructor so will never actually have this value.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        $response = $this->httpClient->request($method, $url, $body);
+        $statusCode = $response->getStatusCode();
+        if (200 === $statusCode) {
+            $content = json_decode($response->getContent());
+
+            if (is_null($content)) {
+                $message = $this->getErrorMessage(0);
+                throw new OcrException('transkribus-error', [$message]);
```
samwilson: The `transkribus-error` message needs to be defined in `en.json` and `qqq.json`.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+            'image' => [
+                'imageUrl' => $imageURL,
+            ],
+        ];
+
+        $processId = 0;
+        $content = $this->request('POST', self::PROCESSES_URL, $jsonBody);
+        $parsedContent = (array) $content;
+        if (!empty($parsedContent)) {
+            if ($content->{'status'} === 'CREATED') {
```
samwilson: There are five places where you're referring to member variables with braces where these don't contain illegal characters. These can all be simplified to e.g. `$content->status`.

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * @param int $processId
+     * @return string
+     */
+    public function retrieveProcessResult(int $processId): string
+    {
+        $url = self::PROCESSES_URL.'/'.$processId;
+
+        $content = $this->request('GET', $url, []);
+        $parsedContent = (array) $content;
```
samwilson: Why are you casting to an array?

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+        $textResult = '';
+
+        if (!empty($parsedContent)) {
+            if ($content->{'status'} === 'FINISHED') {
+                $textContent = $content->{'content'};
+                $textResult = $textContent->{'text'};
```
samwilson: This could be simplified to `$textResult->content->text ?? null;`

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+    public function retrieveProcessResult(int $processId): string
+    {
+        $url = self::PROCESSES_URL.'/'.$processId;
+
+        $content = $this->request('GET', $url, []);
+        $parsedContent = (array) $content;
+        $textResult = '';
+
+        if (!empty($parsedContent)) {
+            if ($content->{'status'} === 'FINISHED') {
```
samwilson: How are you distinguishing between a finished process and one that returns an empty string? If we call `retrieveProcessResult()` and get an empty string, we don't know whether to call it again or not.

src/Engine/TranskribusEngine.php
```diff
@@ -0,0 +1,69 @@
...
+        array $crop,
+        ?array $langs = null
+    ): EngineResult {
+        $this->checkImageUrl($imageUrl);
+
+        $image = $this->getImage($imageUrl, $crop);
+        $imageUrl = $image->getUrl();
+        $processId = $this->transkribusClient->initProcess($imageUrl);
+
+        if (0 === $processId) {
```
samwilson: The process ID of `0` is hard coded in the client. I suggest that initProcess instead should throw an exception if no process ID was returned.

templates/output.html.twig
```diff
@@ -23,6 +23,12 @@
                         {{ msg('engine-name-tesseract') }}
                     </label>
                 </div>
+                <div class="radio">
+                    <label>
+                        <input type="radio" name="engine" value="transkribus" {% if engine == 'transkribus' %}checked{% endif %} />
+                        {{ msg('engine-name-transkribus') }}
+                    </label>
+                </div>
```
samwilson, line 31: This UI is pretty simple, so it's good to have it here. But note that this task is about adding to the API so should also have additional edits to the OpenApi docs for the `apiAction()` method in the controller.

Parthiv-M (approved):

src/Engine/TranskribusClient.php
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * @param int $statusCode
+     * @return string
+     */
+    private function getErrorMessage(int $statusCode): string
+    {
+        $errorMessage = '';
+        switch ($statusCode) {
+            case 0:
+                $errorMessage = 'Transkribus API returned null';
```
Parthiv-M: Should we be changing this message to something like `Could not parse result from Transkribus API` instead?

Parthiv-M (changes requested):

src/Engine/TranskribusClient.php
kolakachi: Noted, I think we need a parameter for content. This holds image co-ordinates.

samwilson: That's true, but it's not being used yet so there's no need to add it yet. It can be refactored as required.

src/Engine/TranskribusClient.php
kolakachi: I noticed The API has different error messages and status codes, is it advisable to let Tranksribus define the error messages? or do we go along with defining the messages ourselves?

src/Engine/TranskribusClient.php
kolakachi: Currently checking for FINISHED and FAILED. The code currently covers CREATED, WAITING and RUNNING.

src/Controller/OcrController.php
```diff
@@ -184,7 +185,7 @@ public function homeAction(): Response
      * @OA\Parameter(
      *     name="engine",
      *     in="query",
-     *     description="The engine to use, either `tesseract` or `google`.",
+     *     description="The engine to use, either `tesseract` or `google` or transkribus.",
```
Parthiv-M: I think it should be changed to \`transkribus`, so that it conforms to the rest of the engine name styles

Parthiv-M:
Also, the `description` field [here](https://github.com/wikimedia/wikimedia-ocr/blob/2b7ec341a505f23c04e66650053a46cf614a4d1d/config/packages/nelmio_api_doc.yaml#L5) also needs to be changed to include Transkribus

Parthiv-M (approved):

i18n/qqq.json
```diff
@@ -56,5 +57,10 @@
 	"tesseract-psm-12": "Form option for Tesseract page segmentation mode.",
 	"tesseract-psm-13": "Form option for Tesseract page segmentation mode.",
 	"tesseract-param-error": "Error message displayed when invalid values for Tesseract options are submitted.\n\nParameters:\n* $1 – the form label for the option. Currently, the only possible value here is {{msg-wm|wikimedia-ocr-tesseract-psm-label}}.\n* $2 – The value that was given.\n* $3 – the maximum value for the option (this will be an integer).",
-	"tesseract-internal-error": "Generic error message displayed when the tesseract command fails."
+	"tesseract-internal-error": "Generic error message displayed when the tesseract command fails.",
+	"transkribus-error-401": "Error Code '$1' :: The request is not authorized",
```
samwilson: The sentences in qqq.json are the help text to go along with the actual messages in en.json. They're shown to translators on TranslateWiki.net, who will use them to make decisions about how to translate things. So the need to explain what the message is for. See https://www.mediawiki.org/wiki/Help:System_message for some more info (this tool uses the pretty much the same system as MediaWiki).

samwilson (commented):
Sorry, I was just about to merge this but I noticed that the message documentation isn't correct. (As soon as patches are merged, their i18n data is updated on translatewiki.net, so we can't do this in a follow-up commit.)

PR #67, opened 2023-02-20
Add UI config for transkribus engine

Parthiv-M:
Added model and region selection for the Transkribus engine

Bug: T326524

### DIFF
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -39,6 +39,10 @@ abstract class EngineBase
     public const LANG_NAMES = [
         'az-cyrl' => 'Azərbaycan (qədim yazı)',
         'de-frk' => 'Deutsch (Fraktur)',
+        'en-18-handwritten' => 'English Handwritten (18-19th Century)',
+        'en-print-m1' => 'English Typewritten/Print',
+        'en-print-03' => 'English Typewritten/Print',
+        'en-typewriter' => 'English Typewritten',
         'enm' => 'Middle English (1100-1500)',
         'es-old' => 'español (viejo)',
         'frm' => 'moyen français (1400-1600)',
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -39,10 +39,8 @@ abstract class EngineBase
     public const LANG_NAMES = [
         'az-cyrl' => 'Azərbaycan (qədim yazı)',
         'de-frk' => 'Deutsch (Fraktur)',
-        'en-18-handwritten' => 'English Handwritten (18-19th Century)',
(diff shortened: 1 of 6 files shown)


src/Engine/EngineBase.php
```diff
@@ -39,6 +39,10 @@ abstract class EngineBase
     public const LANG_NAMES = [
         'az-cyrl' => 'Azərbaycan (qədim yazı)',
         'de-frk' => 'Deutsch (Fraktur)',
+        'en-18-handwritten' => 'English Handwritten (18-19th Century)',
+        'en-print-m1' => 'English Typewritten/Print',
+        'en-print-03' => 'English Typewritten/Print',
```
kolakachi: The values of `en-print-m1` and `en-print-03` are the same, can we make them different?

src/Engine/TranskribusEngine.php
```diff
@@ -53,15 +53,38 @@ public function getResult(
...
+        if ($validLangs) {
+            $langCodes = $this->getLangCodes($validLangs);
+            if (1 < count($langCodes)) {
+                throw new OcrException('transkribus-multiple-lang-error');
+            } else {
+                $modelId = (int) $langCodes[0];
+            }
+        } else {
+            throw new OcrException('transkribus-no-lang-error');
+        }
```
kolakachi: can we take out the else blocks
```
if (!$validLangs) {
   //throw error
}
$langCodes = $this->getLangCodes($validLangs);
if (1 < count($langCodes)) {
//throw error
}
$modelId = (int) $langCodes[0];
```

src/Engine/EngineBase.php
Parthiv-M: I'm not sure how to make them different. These two models do the same thing, but are named differently, have different CERs and cater to different number of languages. @samwilson what do you suggest?

src/Engine/TranskribusEngine.php
Parthiv-M: Yes sure

src/Engine/EngineBase.php
samwilson: I agree that having them the same is confusing. A user will just see two things that are identical, and have no way to choose between them.

I suggest adding the model names, prefixed with 'Transkribus', e.g.

* Transkribus Balinese balmleaf manuscripts
* Transkribus English Handwriting M3
* Transkribus B2022 English Model M4
* …

I also don't think we should include any Transkribus print models, because we don't want people using this for printed works (due to the cost).

kolakachi (approved):

samwilson:
Sorry! I forgot to include the bug number in the commit message.

PR #68, opened 2023-02-21
Modify auth flow for transkribus to include refresh token

Parthiv-M:
Modify the authentication flow of the Transkribus engine to use refresh tokens in addition to access tokens

Bug: T330052

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -48,8 +48,6 @@ services:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
             $refreshToken: '%env(APP_TRANSKRIBUS_REFRESH_TOKEN)%'
 
-    App\Engine\TranskribusClient:
-
     App\EventListener\ExceptionListener:
         arguments:
             - '@request_stack'
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -44,6 +44,8 @@ services:
 
     App\Engine\TranskribusEngine:
         parent: App\Engine\EngineBase
+
+    App\Engine\TranskribusClient:
         arguments:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
             $refreshToken: '%env(APP_TRANSKRIBUS_REFRESH_TOKEN)%'
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -44,11 +44,11 @@ services:
 
     App\Engine\TranskribusEngine:
         parent: App\Engine\EngineBase
-
-    App\Engine\TranskribusClient:
         arguments:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
             $refreshToken: '%env(APP_TRANSKRIBUS_REFRESH_TOKEN)%'
+
+    App\Engine\TranskribusClient:
 
     App\EventListener\ExceptionListener:
         arguments:
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -48,6 +48,7 @@ services:
     App\Engine\TranskribusClient:
         arguments:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
+            $refreshToken: '%env(APP_TRANSKRIBUS_REFRESH_TOKEN)%'
 
     App\EventListener\ExceptionListener:
         arguments:
--- a/src/Command/TranskribusAuthCommand.php
+++ b/src/Command/TranskribusAuthCommand.php
@@ -90,16 +90,11 @@ class TranskribusAuthCommand extends Command
                         $accessToken = $content->{'access_token'};
                         $refreshToken = $content->{'refresh_token'};
                         $io->writeln('');
-                        $io->writeln('Your access token is: '.$accessToken);
+                        $io->writeln('--- Please copy the following lines and add it to your .env.local ---');
                         $io->writeln('');
-                        $io->writeln('--- Please copy and add it to your .env.local'.
-                                                ' as APP_TRANSKRIBUS_ACCESS_TOKEN ---');
-
+                        $io->writeln('APP_TRANSKRIBUS_ACCESS_TOKEN='.$accessToken);
                         $io->writeln('');
-                        $io->writeln('Your refresh token is: '.$refreshToken);
-                        $io->writeln('');
-                        $io->writeln('--- Please copy and add it to your .env.local'.
-                                                ' as APP_TRANSKRIBUS_REFRESH_TOKEN ---');
+                        $io->writeln('APP_TRANSKRIBUS_REFRESH_TOKEN='.$refreshToken);
                         return Command::SUCCESS;
                     }
                 } elseif (400 === $statusCode) {
--- a/src/Command/TranskribusAuthCommand.php
+++ b/src/Command/TranskribusAuthCommand.php
@@ -13,24 +13,18 @@ use Symfony\Contracts\HttpClient\HttpClientInterface;
 class TranskribusAuthCommand extends Command
 {
 
-    /** @var HttpClientInterface The interface to make http requests */
-    private $client;
-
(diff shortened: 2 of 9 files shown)


src/Command/TranskribusAuthCommand.php
```diff
@@ -88,11 +88,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
                         $io->writeln('Your access token is: '.$accessToken);
                         $io->writeln('');
                         $io->writeln('--- Please copy and add it to your .env.local'.
                                                 ' as APP_TRANSKRIBUS_ACCESS_TOKEN ---');
+
+                        $io->writeln('');
+                        $io->writeln('Your refresh token is: '.$refreshToken);
+                        $io->writeln('');
+                        $io->writeln('--- Please copy and add it to your .env.local'.
+                                                ' as APP_TRANSKRIBUS_REFRESH_TOKEN ---');
```
samwilson: It might be good to output exactly what needs to be copied into `.env.local`, so we don't have to copy and paste twice.

src/Engine/TranskribusClient.php
```diff
@@ -137,8 +146,52 @@ private function request(
...
+                }
+                $this->throwException(0);
+            }
+        }
+        $this->throwException($statusCode);
+    }
+
+    private function refreshAccessToken(): void
+    {
+        $response = $this->httpClient->request(
```
samwilson, line 222: This request is identical to the one done in `TranskribusAuthCommand` isn't it? I wonder if it'd be neater to inject a `TranskribusClient` into the command, and so only have this code in one place?

tests/OcrTestCase.php
```diff
@@ -13,10 +13,14 @@ class OcrTestCase extends KernelTestCase
...
 
+    /** @var string */
+    protected $refreshToken;
+
     protected function setUp(): void
     {
         self::bootKernel();
         $this->projectDir = self::$kernel->getProjectDir();
         $this->accessToken = "eyJhbGciOiJSUzI1NiIsInR5cC";
+        $this->refreshToken = "eyJhbGciOiJSUzI1NiIsInR5cC";
```
samwilson: Sorry, I know I should look this up, but shouldn't test env vars like these be defined in `phunit.xml.dist`? If not, no worries.

Also, it might be good to make these different values, and comprehensible, so people don't think they're meant to be identical or useful. e.g. `eyThisIsATestAccessToken`.

src/Engine/TranskribusClient.php
kolakachi: New static methods have been added.

tests/OcrTestCase.php
samwilson: Do they not work when added to `phpunit.xml.dist`?

src/Engine/TranskribusClient.php
samwilson: The method shouldn't be static. A TranskribusClient object should be injected in the constructor of the command, and then the method used non-statically. Does that make sense?

src/Engine/TranskribusClient.php
```diff
@@ -137,8 +146,97 @@ private function request(
...
+     * @param string $userName
+     * @param string $password
+     * @param HttpClientInterface $client
+     */
+    public static function getAccessTokenResponse(
+        string $userName,
+        string $password,
+        HttpClientInterface $client
+    ): object {
+        $body = [
```
Parthiv-M: I believe this function is not being called anywhere. Can we remove it?

kolakachi: The Intent is for it to be used in the TranskribusAuthCommand class to have the requests in one place

src/Engine/TranskribusClient.php
kolakachi: Static methods to avoid injecting empty accessToken/refereshToken when running the command for the first time. Updates have been made. The methods are no longer static. The client no longer requires the accesstoken and refreshtoken when it's created.

tests/Controller/OcrControllerTest.php
```diff
@@ -47,7 +47,9 @@ public function testGetLang(array $getParams, array $expectedLangs): void
                 $gcv,
                 new TesseractEngine(new MockHttpClient(), $intuition, $this->projectDir, new TesseractOCR()),
                 new TranskribusEngine(
-                    new TranskribusClient(new MockHttpClient(), $this->accessToken),
+                    new TranskribusClient(new MockHttpClient()),
+                    $_ENV['APP_TRANSKRIBUS_ACCESS_TOKEN'],
+                    $_ENV['APP_TRANSKRIBUS_REFRESH_TOKEN'],
```
samwilson: These should use `getenv()`, to avoid use of globals.

config/services.yaml
```diff
@@ -44,10 +44,11 @@ services:
 
     App\Engine\TranskribusEngine:
         parent: App\Engine\EngineBase
-
-    App\Engine\TranskribusClient:
         arguments:
             $accessToken: '%env(APP_TRANSKRIBUS_ACCESS_TOKEN)%'
+            $refreshToken: '%env(APP_TRANSKRIBUS_REFRESH_TOKEN)%'
+
+    App\Engine\TranskribusClient:
```
samwilson: Is this empty config needed?

src/Command/TranskribusAuthCommand.php
```diff
@@ -15,19 +16,37 @@ class TranskribusAuthCommand extends Command
     /** @var HttpClientInterface The interface to make http requests */
     private $client;
```
samwilson: This is no longer used.

src/Command/TranskribusAuthCommand.php
```diff
@@ -15,19 +16,37 @@ class TranskribusAuthCommand extends Command
...
+    /** @var TranskribusClient The TranskribusClient used to make authorization calls */
+    private $transkribusClient;
+
+    /**
+     * TranskribusAuthCommand constructor.
+     * @param HttpClientInterface $client
+     * @param TranskribusClient $transkribusClient
+     */
+    public function __construct(
+        HttpClientInterface $client,
```
samwilson: This service can be removed now, I think?

tests/OcrTestCase.php
```diff
@@ -13,10 +13,12 @@ class OcrTestCase extends KernelTestCase
     /** @var string */
     protected $accessToken;
 
+    /** @var string */
+    protected $refreshToken;
```
samwilson: This is not used (and $accessToken can also be removed).

src/Engine/TranskribusClient.php
```diff
@@ -137,8 +149,89 @@ private function request(
...
+            $response = $this->httpClient->request($method, $url, $body);
+            $statusCode = $response->getStatusCode();
+
+            if (200 === $statusCode) {
+                $content = json_decode($response->getContent());
+
+                if (!empty($content)) {
+                    return $content;
+                }
+                $this->throwException(0);
```
samwilson: I know it's not being introduced in this patch, but these magic numbers are not great I think. Better to set the codes to be class constants and then use them here, e.g. `$this->throwException(self::ERROR_NO_CONTENT);` (can be done in a separate patch of course).

samwilson (commented):
Looks great, works in my tests. Just a few errant variables that are no longer required, but otherwise good to go I think.

kolakachi (approved):

src/Engine/TranskribusEngine.php
```diff
@@ -22,13 +22,16 @@ class TranskribusEngine extends EngineBase
...
+        string $accessToken,
+        string $refreshToken,
         Intuition $intuition,
         string $projectDir,
         HttpClientInterface $httpClient
     ) {
         parent::__construct($intuition, $projectDir, $httpClient);
 
         $this->transkribusClient = $transkribusClient;
+        $this->transkribusClient->setTokens($accessToken, $refreshToken);
```
samwilson: The tokens are only injected here in order to be passed through to the client — would it be better to inject them into the client instead?

PR #70, opened 2023-03-01
Add request model option for Transkribus engine

Parthiv-M:
Add links to allow users to request for a model from the list of public models available for Transkribus

Bug: T330642

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -58,6 +58,6 @@
 	"transkribus-failed-process-error": "Transkribus process failed",
     "transkribus-no-lang-error": "No language was selected",
     "transkribus-multiple-lang-error": "Multiple languages are not allowed, specify one language",
-    "transkribus-browse-public-models": "Browse all public models for Transkribus",
+    "transkribus-browse-public-models": "Browse all public language models for Transkribus",
     "transkribus-request-for-model": "Make a request to add a model from Transkribus to the OCR tool"
 }
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -57,5 +57,7 @@
 	"transkribus-init-process-error": "Failed to initialize Transkribus process",
 	"transkribus-failed-process-error": "Transkribus process failed",
     "transkribus-no-lang-error": "No language was selected",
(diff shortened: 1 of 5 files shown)


templates/_transkribus_help.html.twig
```diff
@@ -0,0 +1,5 @@
+<div id="transkribus-options" class="engine-options engine-help {% if engine != 'transkribus' %}hidden{% endif %}">
+	<a href="https://readcoop.eu/transkribus/public-models">{{ msg('transkribus-browse-public-models') }}</a>
+	<br />
+	<a href="https://w.wiki/6PJ5" target="_blank">{{ msg('transkribus-request-for-model') }}</a>
```
samwilson: I think it'd be okay to have the whole Phabricator URL here, rather than use the shortener. It'd be easier to read, and be searchable in codesearch. The [url_encode](https://twig.symfony.com/doc/3.x/filters/url_encode.html) Twig filter can be used.

Also, it doesn't need to list us all as subscribers because by adding the project tag we'll all get notified anywhere (because we're watchers of the project).

i18n/en.json
```diff
@@ -57,5 +57,7 @@
 	"transkribus-init-process-error": "Failed to initialize Transkribus process",
 	"transkribus-failed-process-error": "Transkribus process failed",
     "transkribus-no-lang-error": "No language was selected",
-    "transkribus-multiple-lang-error": "Multiple languages are not allowed, specify one language"
+    "transkribus-multiple-lang-error": "Multiple languages are not allowed, specify one language",
+    "transkribus-browse-public-models": "Browse all public models for Transkribus",
```
samwilson: I know we've T330061 open to sort this out, but I think that having it say "public models" here when the field above is "Languages" might be confusing. Users wouldn't know what a model is at this point. Maybe just say "public language models" here; what do you think? @satdeepgill what do you think?

templates/_transkribus_help.html.twig
Parthiv-M: Alright

i18n/en.json
satdeepgill: I agree! “public language models” would be better understood.

PR #71, opened 2023-03-02
Change API error for invalid langs to JSON

kolakachi:
Added a try and catch block in apiAction in OcrController, Catch block returns a json response.

Bug: T330243

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -10,6 +10,7 @@ use App\Engine\GoogleCloudVisionEngine;
 use App\Engine\TesseractEngine;
 use App\Engine\TranskribusEngine;
 use App\Exception\EngineNotFoundException;
+use Exception;
 use Krinkle\Intuition\Intuition;
 use OpenApi\Annotations as OA;
 use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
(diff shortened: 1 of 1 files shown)


Parthiv-M (approved):

Parthiv-M (commented):
Worked fine for me

src/Controller/OcrController.php
```diff
@@ -220,16 +221,22 @@ public function homeAction(): Response
...
+                'text' => $result->getText(),
+            ]);
+            $warnings = $result->getWarnings();
+            if ($warnings) {
+                $responseParams['warnings'] = $warnings;
+            }
+            return $this->getApiResponse($responseParams);
+        } catch (Exception $exception) {
+            return $this->getApiResponse([
+                "message" => $exception->getMessage(),
```
samwilson: Other error states return the error in the `error` key, so this should probably do the same, rather than `message`.

src/Controller/OcrController.php
```diff
@@ -220,16 +221,22 @@ public function homeAction(): Response
...
-        $this->setup();
-        $result = $this->getResult(EngineBase::WARN_ON_INVALID_LANGS);
-        $responseParams = array_merge(static::$params, [
-            'text' => $result->getText(),
-        ]);
-        $warnings = $result->getWarnings();
-        if ($warnings) {
-            $responseParams['warnings'] = $warnings;
+        try {
+            $this->setup();
```
samwilson, line 225: The exception is actually only coming from this method call isn't it? It might be better to just try/catch around this then, because the other calls below already have functioning error handling.

samwilson (changes requested):
It looks like this breaks the error output of requests such as `/api?langs[]=foobar`, which now returns only `{"message":""}` instead of the error message and other info.

PR #72, opened 2023-03-10
Keep selected languages when changing engine

kolakachi:
Updated app.js to keep only the supported languages when switching back to another engine

Bug: T329666


samwilson (commented):
This looks good.

Is the intention with using localStorage that the lang list be persisted between page reloads? Because it looks at the moment like it doesn't do this. If it's just for storing the lang list while changing the radio buttons, then it doesn't need to be localStorage but can instead be a normal variable.

PR #73, opened 2023-03-11
Make languages field mandatory for Transkribus engine

Parthiv-M:
Modify dropdown field and label to mark the languages input as _required_ for the Transkribus engine

Bug: T330829

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -53,6 +53,9 @@ $(function () {
         placeholder: $select2.data('placeholder'),
     });
 
+    let selectLangField = $('#lang');
+    let previousDataPlaceholder = selectLangField.attr('data-placeholder');
+
     // Show engine-specific options.
     $('[name=engine]').on('change',  e => {
         updateSelect2Options(e.target.value);
@@ -60,11 +63,13 @@ $(function () {
         $(`#${e.target.value}-options`).removeClass('hidden');
         let engine = e.target.value;
         if(engine === 'tesseract' || engine === 'google') {
-            $('#lang').prop('required', false);
+            selectLangField.prop('required', false);
+            selectLangField.attr('data-placeholder', previousDataPlaceholder);
             $('#transkribus-lang-label').addClass('hidden');
             $('#optional-lang-label').removeClass('hidden');
         } else {
-            $('#lang').prop('required', true);
+            selectLangField.prop('required', true);
+            selectLangField.attr('data-placeholder', '');
             $('#optional-lang-label').addClass('hidden');
             $('#transkribus-lang-label').removeClass('hidden');
         }
--- a/assets/app.js
+++ b/assets/app.js
@@ -58,6 +58,14 @@ $(function () {
         updateSelect2Options(e.target.value);
         $('.engine-options').addClass('hidden');
         $(`#${e.target.value}-options`).removeClass('hidden');
+        let engine = e.target.value;
+        if(engine === 'tesseract' || engine === 'google') {
+            $('#lang').removeAttr('required');
+            $(`#optional-lang-label`).removeClass('hidden');
+        } else {
+            $('#lang').attr('required', '');
+            $(`#transkribus-lang-label`).removeClass('hidden');
+        }
     });
 
     // For the result page. Makes the 'Copy' button copy the transcription to the clipboard.
--- a/assets/app.js
+++ b/assets/app.js
@@ -91,6 +91,14 @@ $(function () {
         }
     });
 
+    // modify selected engine after loading the page with preselected engine 
+    let engineRadioFields = $('[name=engine]:checked');
+    if(engineRadioFields.val() === 'transkribus') {
+        selectLangField.attr('data-placeholder', '')
+    } else {
+        selectLangField.attr('data-placeholder', previousDataPlaceholder);
+    }
+
     // For the result page. Makes the 'Copy' button copy the transcription to the clipboard.
     const $copyButton = $('.copy-button');
     if ($copyButton.length) {
--- a/assets/app.js
+++ b/assets/app.js
@@ -69,8 +69,7 @@ $(function () {
         });
     });
 
-    let selectLangField = $('#lang');
-    let previousDataPlaceholder = selectLangField.attr('data-placeholder');
+    let previousDataPlaceholder = $select2.attr('data-placeholder');
 
     // Show engine-specific options.
     $('[name=engine]').on('change',  e => {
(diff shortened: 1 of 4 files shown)


kolakachi (approved):

assets/app.js
```diff
@@ -58,6 +58,14 @@ $(function () {
         updateSelect2Options(e.target.value);
         $('.engine-options').addClass('hidden');
         $(`#${e.target.value}-options`).removeClass('hidden');
+        let engine = e.target.value;
+        if(engine === 'tesseract' || engine === 'google') {
+            $('#lang').removeAttr('required');
```
samwilson: I think valueless properties are better off using `.prop( 'required', false )` to avoid issues in some browsers.

assets/app.js
```diff
@@ -58,6 +58,14 @@ $(function () {
...
+            $(`#optional-lang-label`).removeClass('hidden');
```
samwilson: Is this using backticks intentionally? It doesn't contain any variables. Same with the one below.

assets/app.js
```diff
@@ -58,6 +58,14 @@ $(function () {
...
+        } else {
+            $('#lang').attr('required', '');
+            $(`#transkribus-lang-label`).removeClass('hidden');
+        }
```
samwilson, line 92: Each branch of this conditional should also `addClass('hidden')` on the other `<label>`, otherwise both end up being displayed at the same time (which looks odd).

templates/output.html.twig
```diff
@@ -31,7 +31,12 @@
                 </div>
             </div>
             <div class="form-group">
-                <label for="lang">{{ msg('language-code') }}</label>
+                <label id="optional-lang-label" for="lang" class="lang-label {% if engine == 'transkribus' %}hidden{% endif %}">
+                    {{ msg('language-code') }}
+                </label>
+                <label id="transkribus-lang-label" for="lang" class="lang-label {% if engine != 'transkribus' %}hidden{% endif %}">
+                    {{ msg('transkribus-language-code') }}
+                </label>
```
samwilson, line 39: The #lang select below doesn't get `required` on first load (i.e. when `?engine=transkribus` is given in the URL).

assets/app.js
```diff
@@ -58,6 +58,16 @@ $(function () {
         updateSelect2Options(e.target.value);
         $('.engine-options').addClass('hidden');
         $(`#${e.target.value}-options`).removeClass('hidden');
+        let engine = e.target.value;
+        if(engine === 'tesseract' || engine === 'google') {
+            $('#lang').removeAttr('required');
```
samwilson: `.prop('required', false)` should be used to unset the `required` attribute, see https://api.jquery.com/prop/#prop2

templates/output.html.twig
```diff
@@ -31,8 +31,13 @@
...
             <div class="form-group">
-                <label for="lang">{{ msg('language-code') }}</label>
-                <select id="lang" class="form-control" name="langs[]" multiple data-placeholder="{{ msg('langs-placeholder') }}">
+                <label id="optional-lang-label" for="lang" class="lang-label {% if engine == 'transkribus' %}hidden{% endif %}">
+                    {{ msg('language-code') }}
+                </label>
+                <label id="transkribus-lang-label" for="lang" class="lang-label {% if engine != 'transkribus' %}hidden{% endif %}">
+                    {{ msg('transkribus-language-code') }}
+                </label>
+                <select id="lang" class="form-control" name="langs[]" multiple data-placeholder="{{ msg('langs-placeholder') }}" {% if engine == 'transkribus' %}required=''{% endif %}>
```
samwilson: `required` doesn't need to take a value, it's a boolean attribute (same as `multiple` which is already here).

samwilson (commented):
Works now! Just a couple of minor comments.

templates/output.html.twig
```diff
@@ -31,8 +31,13 @@
...
             <div class="form-group">
-                <label for="lang">{{ msg('language-code') }}</label>
-                <select id="lang" class="form-control" name="langs[]" multiple data-placeholder="{{ msg('langs-placeholder') }}">
+                <label id="optional-lang-label" for="lang" class="lang-label {% if engine == 'transkribus' %}hidden{% endif %}">
+                    {{ msg('language-code') }}
+                </label>
+                <label id="transkribus-lang-label" for="lang" class="lang-label {% if engine != 'transkribus' %}hidden{% endif %}">
+                    {{ msg('transkribus-language-code') }}
+                </label>
+                <select id="lang" class="form-control" name="langs[]" multiple data-placeholder="{{ msg('langs-placeholder') }}" {% if engine == 'transkribus' %}required{% endif %}>
```
samwilson: Sorry, I've just realised that `langs-placeholder` is "Leave blank for automatic language detection." which isn't correct if it's a required field. Probably it can just be made blank for Transkribus, rather than adding a new message to be translated (but do whatever you feel is best).

assets/app.js
```diff
@@ -69,13 +69,36 @@ $(function () {
...
         $(`#${e.target.value}-options`).removeClass('hidden');
+        let engine = e.target.value;
+        if(engine === 'tesseract' || engine === 'google') {
+            selectLangField.prop('required', false);
+            selectLangField.attr('data-placeholder', previousDataPlaceholder);
+            $('#transkribus-lang-label').addClass('hidden');
+            $('#optional-lang-label').removeClass('hidden');
+        } else {
+            selectLangField.prop('required', true);
+            selectLangField.attr('data-placeholder', '');
```
samwilson: Is this blanking the placeholder for you? When I test, it's not making any change.

It looks like it might have to use something like this:

```js
$select2.data('select2').selection.placeholder.text = '';
```

assets/app.js
```diff
@@ -69,13 +69,36 @@ $(function () {
         });
     });
 
+    let selectLangField = $('#lang');
```
samwilson: It looks like `$select2` is already this same selector. Also, note that the convention is to prefix jQuery objects with `$`.

assets/app.js
```diff
@@ -69,13 +69,36 @@ $(function () {
...
+        }
     });
 
+    // modify selected engine after loading the page with preselected engine 
+    let engineRadioFields = $('[name=engine]:checked');
+    if(engineRadioFields.val() === 'transkribus') {
+        selectLangField.attr('data-placeholder', '')
+    } else {
+        selectLangField.attr('data-placeholder', previousDataPlaceholder);
+    }
```
samwilson, line 101: Rather than repeating these here, it might instead be possible to call `.trigger('change')` after the above `.on('change')` and so re-use the same code. But it's fine like this too, I'm sure.

samwilson (commented):
The placeholder switching still isn't working for me, sorry!

PR #78, opened 2023-04-25
Remove phpcs exclusions for linting

Parthiv-M:
Remove `phpcs` exclusions and fix linting errors according to Mediawiki Coding Rules

Bug: T328730

### DIFF
--- a/src/Command/TranskribusAuthCommand.php
+++ b/src/Command/TranskribusAuthCommand.php
@@ -37,6 +37,7 @@ class TranskribusAuthCommand extends Command {
 	 * Method that runs the CLI command.
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
+	 * @return int
 	 */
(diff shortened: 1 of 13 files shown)


src/Command/TranskribusAuthCommand.php
```diff
@@ -37,6 +37,7 @@ protected function configure(): void {
 	 * Method that runs the CLI command.
 	 * @param InputInterface $input
 	 * @param OutputInterface $output
+	 * @return int
```
kolakachi, line 40: The method still returns int

Parthiv-M: I believe `Command::SUCCESS` or `Command::FAILURE` return integers (as mentioned [in the docs](https://symfony.com/doc/current/console.html#creating-a-command))

samwilson (approved):

PR #79, opened 2023-04-30
Disable submit button during form submit

PeterWarrington:
A proposed fix for [Phabricator bug T334161 (Disable "Transcribe Full Page" button during transcription)](https://phabricator.wikimedia.org/T334161) which disables the button when the `onsubmit` event is fired. See the bug report itself for more info.

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -112,9 +112,9 @@ $(function () {
         });
     }
 
-    // Disable submit button during form submit
+    // Disable submit buttons during form submit
     $('form').on('submit', e => {
-        $('.submit-btn .btn').attr("disabled", true);
+        $('.submit-btn .btn, .submit-crop').attr("disabled", true);
     });
 
     var $ocrOutputDiv = $('.ocr-output');
--- a/assets/app.js
+++ b/assets/app.js
@@ -113,9 +113,14 @@ $(function () {
     }
 
     // Disable submit buttons during form submit
+    var submitBtns = $('.submit-btn .btn, .submit-crop')
     $('form').on('submit', e => {
-        $('.submit-btn .btn, .submit-crop').attr("disabled", true);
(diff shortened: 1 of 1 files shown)


kolakachi (commented):
We also have an input type of submit for the button Transcribe area. Can you disable it too on submit?

PeterWarrington:
> We also have an input type of submit for the button Transcribe area. Can you disable it too on submit?

Done - also made buttons re-enable on `pagehide` so that the buttons are re-enabled if the page is returned to via back button, etc.

Parthiv-M (changes requested):
A couple of things I encountered:
- After I click on the _Transcribe whole page_ button and a response is returned, the _Transcribe whole page_ button is re-enabled, but the _Transcribe area_ button still stays disabled. Is that intended behaviour?
- Once I select an area using the crop tool and click on _Transcribe area_ button, the _Transcribe whole page_ button gets disabled on submit but the _Transcribe area_ button remains enabled throughout.

PeterWarrington:
@Parthiv-M 
The transcribe area button is intended to be disabled until a crop area is selected. 
I cannot replicate the other behaviour you describe locally in both Chrome and Firefox - both buttons disable themselves on submit for me. Can you let me know what browser environment you are testing in and verify that you are using the latest commits in this pull request?

Parthiv-M:
I tested with Firefox and have the latest commits from this PR. It does not seem to disable both the buttons when I click on _Transcribe area_.

Parthiv-M:
Hi @PeterWarrington, have you been able to resolve the issue?

Parthiv-M:
Hello @PeterWarrington , will you be making a patch soon for the above mentioned changes, or can we take over and complete the patch if you are caught up with something else?

PR #81, opened 2023-07-15
Retry image-fetching once if the first request is 503

samwilson:
Move the image-fetching and response handling to its own method and call that method recursively once if the first response is a HTTP 503.

Bug: T341916


Parthiv-M (approved):
Looks alright to me!

PR #82, opened 2023-07-26
Fix API docs, endpoint descriptions, etc.

samwilson:
Remove the Nelmio logo from the API docs.

Move phpcs rule disabling to top of file so that it doesn't appear in the API help docs output.

Also update nelmio/api-doc-bundle package.


Parthiv-M (approved):
Works fine for me

PR #83, opened 2023-07-26
Integrate line detection model into Transkribus engine

Parthiv-M:
Add API routes, helper methods and UI components to integrate line detection models into the Transkribus OCR engine

Bug: [T340837](https://phabricator.wikimedia.org/T340837)

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -6,6 +6,7 @@ const $ = require('jquery');
 const $select2 = $('#lang');
 var selectedLanguages = [];
 const $lineDetectionSelect = $('#line-id');
+var lineModels = null;
 
 const Cropper = require('cropperjs');
 import 'cropperjs/dist/cropper.css';
@@ -44,31 +45,46 @@ function updateSelect2Options(engine)
     });
 }
 
+function fetchLineModelsJSON () {
+    $.getJSON('/api/transkribus/available_line_ids').then(response => {
+        lineModels = response.available_line_ids;
+    });
+}
+
 /**
  * Populate the select input field for line detection model IDs with
  * the line detection model IDs available for the Transkribus engine
  */
 function updateLineModelOptions () {
-    $.getJSON('/api/available_line_ids?engine=transkribus').then(response => {
-        const lineModels = response.available_line_ids;
-        
-        // clear existing selections and options
-        $lineDetectionSelect.val(null).empty().trigger('change');
-
-        // append the None option
-        $lineDetectionSelect.append(new Option("None", "0", true, false));
-        // append the default line detection model as a new option
-        $lineDetectionSelect.append(new Option("Default line detection model", "49272", false, false));
-        
-        // append all other line detection models as options
-        Object.keys(lineModels).forEach(model => {
-            const option = new Option(lineModels[model], model, false, false);
-            $lineDetectionSelect.append(option);
+    const staticOptions = $lineDetectionSelect[0].options;
+    let staticOptionData = []; 
+    staticOptions.forEach(option => {
+        staticOptionData.push({
+            text: option.text,
+            value: option.value
         });
+    });
+
+    // clear existing selections and options
+    $lineDetectionSelect.val(null).empty().trigger('change');
+
+    // append static options
+    staticOptionData.slice(0, 2).forEach(staticOption => {
+        $lineDetectionSelect.append(new Option(staticOption.text, staticOption.value, staticOption.value === null, false));    
+    });
+
+    // append all other line detection models as options
+    Object.keys(lineModels).forEach(model => {
+        const option = new Option(lineModels[model], model, false, false);
+        $lineDetectionSelect.append(option);
     });
 }
 
 $(function () {
+
+    // fetch line detection model data
+    fetchLineModelsJSON();
+
     // Remove nojs class, for styling non-Javascript users.
     $('html').removeClass('nojs');
 
@@ -100,7 +116,6 @@ $(function () {
     $('[name=engine]').on('change',  e => {
         let engine = e.target.value;
         updateSelect2Options(engine);
-        updateLineModelOptions();
         $('.engine-options').addClass('hidden');
         $(`#${engine}-options`).removeClass('hidden');
         if(engine === 'tesseract' || engine === 'google') {
@@ -110,7 +125,8 @@ $(function () {
             $('#transkribus-lang-label').addClass('hidden');
             $('#optional-lang-label').removeClass('hidden');
         } else {
-            $('#transkribus-line-ids').removeClass('hidden');
+            updateLineModelOptions();
+            $('#transkribus-help').removeClass('hidden');
             $select2.prop('required', true);
             $select2.attr('data-placeholder', '');
             $select2.data('select2').selection.placeholder.text = '';
--- a/assets/app.js
+++ b/assets/app.js
@@ -5,6 +5,7 @@ import 'select2';
 const $ = require('jquery');
 const $select2 = $('#lang');
 var selectedLanguages = [];
+const $lineDetectionSelect = $('#line-id');
 
 const Cropper = require('cropperjs');
 import 'cropperjs/dist/cropper.css';
@@ -43,6 +44,30 @@ function updateSelect2Options(engine)
     });
 }
 
+/**
+ * Populate the select input field for line detection model IDs with
+ * the line detection model IDs available for the Transkribus engine
+ */
+function updateLineModelOptions () {
+    $.getJSON('/api/available_line_ids?engine=transkribus').then(response => {
+        const lineModels = response.available_line_ids;
+        
+        // clear existing selections and options
+        $lineDetectionSelect.val(null).empty().trigger('change');
+
+        // append the None option
+        $lineDetectionSelect.append(new Option("None", "0", true, false));
+        // append the default line detection model as a new option
+        $lineDetectionSelect.append(new Option("Default line detection model", "49272", false, false));
+        
+        // append all other line detection models as options
+        Object.keys(lineModels).forEach(model => {
+            const option = new Option(lineModels[model], model, false, false);
+            $lineDetectionSelect.append(option);
+        });
+    });
+}
+
 $(function () {
     // Remove nojs class, for styling non-Javascript users.
     $('html').removeClass('nojs');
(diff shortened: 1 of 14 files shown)


src/Controller/OcrController.php
```diff
@@ -262,6 +279,29 @@ public function apiAvailableLangsAction(): JsonResponse {
 		] );
 	}
 
+	/**
+	 * phpcs:disable MediaWiki.Commenting.FunctionAnnotations.UnrecognizedAnnotation
+	 * @Route("/api/available_line_ids", name="apiLineIds", methods={"GET"})
```
samwilson: As line detection models only ever apply to Transkribus, perhaps this API endpoint could be simplified to e.g. `/api/transkribus/line_ids` and not have the `engine` parameter?

It could also have a `lang` parameter, so it'd only return line IDs for a specific language model, but that's probably not worth it given it's such a small result set.

src/Engine/TranskribusEngine.php
```diff
@@ -12,6 +12,12 @@ class TranskribusEngine extends EngineBase {
 	/** @var TranskribusClient */
 	protected $transkribusClient;
 
+	/** @var int line detection model ID to be used by the Transkribus engine */
+	protected $lineId;
+
+	/** @var int Default value for line detection model ID to be used by Transkribus */
+	public const DEFAULT_LINEID = 0;
```
samwilson, line 19: Does `0` have a specific meaning for this in the Transkribus API? If not, it might be neater to make the default `null` and check instead for that above.

templates/_transkribus_options.html.twig
```diff
@@ -0,0 +1,23 @@
+<fieldset id="transkribus-line-ids" class="engine-options {% if engine != 'transkribus' %}hidden{% endif %}">
```
samwilson: The ID of this element should be `transkribus-options` (in line with the tesseract-options element) but that's already in use in `_transkribus_help.html.twig`. I'd suggest renaming that, so that this gets its rightful name. That will also simplify the JS switching logic.

assets/app.js
```diff
@@ -43,6 +44,30 @@ function updateSelect2Options(engine)
     });
 }
 
+/**
+ * Populate the select input field for line detection model IDs with
+ * the line detection model IDs available for the Transkribus engine
+ */
+function updateLineModelOptions () {
+    $.getJSON('/api/available_line_ids?engine=transkribus').then(response => {
```
samwilson: Does this need to be re-fetched every time the engine selection is changed? It currently returns the same thing every time, so could probably just be done once on page load.

It could also perhaps be sending the language code and so only showing the valid line IDs for the currently-selected language.

templates/_transkribus_options.html.twig
```diff
@@ -0,0 +1,23 @@
...
+    <div class="form-group">
+        <label for="lineId">{{ msg('transkribus-line-label') }}</label>
+        <select
+            name="lineId"
+            id="line-id"
+            class="form-control"
+            data-placeholder="Choose a line detection model"
+        >
+            <option value="0" {% if lineId == 0 %}selected{% endif %}>None</option>
+            <option value="49272" {% if lineId == 49272 %}selected{% endif %}>Default line detection model</option>
```
samwilson: Here and above in `app.js` it's adding these two options, but I'm confused about what the difference is between *not* sending a line ID and sending what we know to be the *default*. This looks like it'd be fine for now (although unnecessary), but if the default ever changes then this will be incorrect. Alternatively, 49272 is not applied unless we send it, but in that case we shouldn't be calling it 'default'.

Also, the text here and for the placeholder should be translated.

src/Controller/OcrController.php
```diff
@@ -165,6 +175,13 @@ public function homeAction(): Response {
...
 
+		// Pre-supply the available line ids for autocompletion in the form.
+		static::$params['available_line_ids'] = $this->engine->getValidLineIds( true, false );
+		sort( static::$params['available_line_ids'] );
+
+		static::$params['available_line_id_langs'] = $this->engine->getValidLineIds( false, true );
+		sort( static::$params['available_line_id_langs'] );
+
 		// Intution::listToText() isn't available via Twig, and we only want to do this for the view and not the API.
 		static::$params['image_hosts'] = $this->intuition->listToText( static::$params['image_hosts'] );
```
samwilson, line 192: Below here in `apiAction()` shouldn't there be a new parameter of `line_id`?

samwilson:
Also, I'm getting a 403 response from Transkribus with the wikimedia account: "Your account does not have access to text recognition model with ID = 45612". How do I fix that? Do you have it working correctly?

src/Engine/TranskribusEngine.php
Parthiv-M: `null` should be fine

templates/_transkribus_options.html.twig
Parthiv-M: Got it

templates/_transkribus_options.html.twig
Parthiv-M: It's more of a semantic error on my part 😅. No model being selected implies we let Transkribus choose the model automatically, whereas using the default model (49272) implies using the _public model that works best in most cases_, as suggested by Transkribus. 

[According to Transkribus](https://transkribus.eu/TrpServer/rest/models/layout/49272), model 49272 is named _Mixed Line Orientation_. Can we go ahead and use the same name?

Parthiv-M:
> Also, I'm getting a 403 response from Transkribus with the wikimedia account: "Your account does not have access to text recognition model with ID = 45612". How do I fix that? Do you have it working correctly?

I believe this 403 response is because the Balinese line detection model is a private model. We might need to reach out to Transkribus regarding that.

templates/_transkribus_options.html.twig
samwilson: Ah that makes sense, and yep that name sounds fine (although I wonder how it'll translate… it's no worse than the PSM names though).

Parthiv-M: Okayy

src/Controller/OcrController.php
Parthiv-M: I just noticed, the `?engine` query is required by the [`setup()` method](https://github.com/wikimedia/wikimedia-ocr/blob/79dc7d0598c33bc18932e78edcccea6e71d34028/src/Controller/OcrController.php#L91) to set the OCR engine for all further operations.

samwilson: Maybe it could set `static::$params['engine']` in this method? As it stands, people could pass e.g. `/api/available_line_ids?engine=google`, which wouldn't really make any sense.

src/Engine/EngineBase.php
```diff
@@ -143,6 +148,46 @@ public function getValidLangs( bool $withNames = false ): array {
 		return $list;
 	}
 
+	/**
+	 * Get line detection models accepted by the engine
+	 * @param bool $onlyLineIds Whether to return only the line detection model IDs
+	 * @param bool $onlyLineIdLangs Whether to return only the line detection model IDs lang codes
+	 * @return string[] Line detection model lang codes or model IDs or model ID names
+	 */
+	public function getValidLineIds( bool $onlyLineIds = false, bool $onlyLineIdLangs = false ): array {
```
samwilson: I'm not sure Line IDs should be part of the EngineBase; they're an attribute just of Transkribus, so shouldn't be at this higher level.

src/Engine/EngineBase.php
```diff
@@ -154,10 +199,19 @@ public function getLangName( ?string $lang = null ): string {
...
+	 * Get name of the given line detection model
+	 * @param string|null $lineIdLang
+	 * @return string
+	 */
+	public function getLineIdModelName( ?string $lineIdLang = null ): string {
+		return self::LINE_ID_MODEL_NAMES[$lineIdLang];
+	}
+
 	/**
 	 * Transform the given ISO 639-1 codes into the language codes needed by this type of Engine.
```
samwilson, line 158: This comment, and perhaps the method name as well, should probably be updated to reflect that the return value is 'language model information' or something, rather than just "the language codes".

src/Controller/OcrController.php
Parthiv-M: Yupp, doing that

src/Engine/EngineBase.php
Parthiv-M: I'm not sure I get you on this. This method just returns the name of the model from the array `LINE_ID_MODEL_NAMES`.

src/Controller/OcrController.php
```diff
@@ -165,6 +175,15 @@ public function homeAction(): Response {
 		static::$params['available_langs'] = $this->engine->getValidLangs();
 		sort( static::$params['available_langs'] );
 
+		if ( static::$params['engine'] === 'transkribus' ) {
+			// Pre-supply the available line ids for autocompletion in the form.
+			static::$params['available_line_ids'] = $this->engine->getValidLineIds( true, false );
```
samwilson, line 183: This array key also needs to be set in `OcrController::$params`, otherwise it throws an error when loading without `?engine=transkribus`:

> Variable "available_line_ids" does not exist.
> in templates/_transkribus_options.html.twig (line 12)

src/Controller/OcrController.php
```diff
@@ -60,6 +60,7 @@ class OcrController extends AbstractController {
 		'langs' => [],
 		'psm' => TesseractEngine::DEFAULT_PSM,
 		'crop' => [],
+		'lineId' => TranskribusEngine::DEFAULT_LINEID
```
samwilson: Nitpick: it doesn't matter at all, but I think URL & API component names should either be camelCase or snake_case, but not both. It looks like we've got more of the latter so far (although most are single words so aren't either), so I suggest `line_id` here.

src/Controller/OcrController.php
```diff
@@ -262,6 +287,22 @@ public function apiAvailableLangsAction(): JsonResponse {
...
+	 * @Route("/api/transkribus/available_line_ids", name="apiLineIds", methods={"GET"})
+	 * OA\Response(response=200, description="List of available line detection model IDs, in JSON format")
+	 * phpcs:enable
+	 * @return JsonResponse
+	 */
+	public function apiAvailableLineDetectionModelIds(): JsonResponse {
+		$this->request->query->set( 'engine', 'transkribus' );
+		$this->setup();
+		return $this->getApiResponse( [
+			'engine' => 'transkribus',
```
samwilson: As `transkribus` is now in the API endpoint URL, there's probably no need to also list it in the returned data.

samwilson (commented):
Looks good, works in my tests. Just one bug with there being no default value.

PR #89, opened 2023-08-19
Add kraken OCR engine

stweil:


stweil:
This is a first implementation to support the Open Source kraken OCR engine.

I created this draft pull request to allow public review and comments although the implementation is still incomplete. OCR for cropped image still needs testing, and there is also currently no unit test code for the new engine.

samwilson:
Also, would you mind opening a Phabricator task for this work, so it can be tracked there? Thanks!

stweil:
> Also, would you mind opening a Phabricator task for this work, so it can be tracked there? Thanks!

Done, see https://phabricator.wikimedia.org/T345055. I also updated the PR here to solve a merge conflict.

stweil:
See also my [test installation](https://kraken-ocr.wmcloud.org/).

stweil:
The test installation is meanwhile available on https://kraken-ocr.wmcloud.org/.

src/Controller/OcrController.php
```diff
@@ -290,6 +334,28 @@ public function apiAvailableLangsAction(): JsonResponse {
 		] );
 	}
 
+	/**
+	 * Get a list of available segmentation models for use with a specific OCR engine.
+	 *
+	 * @Route("/api/available_segmentation_models", name="apiSegmentationModels", methods={"GET"})
```
Parthiv-M, line 340: This route, since it is related to Kraken, should be named `/api/kraken/available_segmentation_models`

public/langs.json
```diff
@@ -265,6 +286,11 @@
             "htr": 45902
         }
     },
+    "german-fraktur-19th-20th-century": {
+        "transkribus": {
+            "htr": 37738
+        }
+    },
```
Parthiv-M, line 293: This Transkribus model should not be added along with Kraken changes, it would be better to separate it out

src/Engine/TranskribusEngine.php
```diff
@@ -118,9 +118,6 @@ public function getResult(
 	): EngineResult {
 		$this->checkImageUrl( $imageUrl );
 
-		$image = $this->getImage( $imageUrl, $crop );
-		$imageUrl = $image->getUrl();
-
 		$points = '';
 		if ( $crop ) {
 			$x = $crop['x'];
```
Parthiv-M, line 123: We'd prefer to isolate changes other than those related to Kraken to another PR!

tests/Engine/EngineBaseTest.php
```diff
@@ -28,11 +32,13 @@ class EngineBaseTest extends OcrTestCase {
...
-		$this->googleEngine = $this->instatiateEngine( 'google' );
+		$this->googleEngine = $this->instantiateEngine( 'google' );
+
+		$this->tesseractEngine = $this->instantiateEngine( 'tesseract' );
 
-		$this->tesseractEngine = $this->instatiateEngine( 'tesseract' );
+		$this->transkribusEngine = $this->instantiateEngine( 'transkribus' );
 
-		$this->transkribusEngine = $this->instatiateEngine( 'transkribus' );
+		$this->krakenEngine = $this->instantiateEngine( 'kraken' );
```
Parthiv-M, line 41: `instantiate()` fixed in #115

config/packages/nelmio_api_doc.yaml
```diff
@@ -2,10 +2,11 @@ nelmio_api_doc:
     documentation:
         info:
             title: Wikimedia OCR
-            description: A web service for Tesseract, Google and Transkribus OCR engines.
-            version: 1.0.0
+            description: A web service for Kraken, Tesseract, Google and Transkribus OCR engines.
+            version: 1.4.0
```
Parthiv-M, line 6: I believe it has been bumped up to 1.4.4 now.

config/packages/nelmio_api_doc.yaml
```diff
@@ -2,10 +2,11 @@ nelmio_api_doc:
...
     areas:
         path_patterns:
             - ^/api$
             - ^/api/available_langs$
+            - ^/api/available_segmentation_models$
```
Parthiv-M, line 11: Will need to change this in accordance with my comment on route path

i18n/en.json
```diff
@@ -1,6 +1,6 @@
 {
     "@metadata": {},
-    "title": "WikimediaOCR",
+    "title": "WikimediaOCR – Kraken Test",
```
Parthiv-M, line 3: This should remain as **WikimediaOCR**

package.json
```diff
@@ -13,6 +13,7 @@
         "regenerator-runtime": "^0.13.11",
         "select2": "^4.0.13",
         "select2-bootstrap-theme": "0.1.0-beta.10",
+        "stylelint": "^15.10.3",
```
Parthiv-M, line 16: `stylelint` can be removed from this PR as well

package.json
```diff
@@ -23,5 +24,6 @@
         "watch": "encore dev --watch",
         "build": "encore production --progress",
         "test": "grunt test"
-    }
+    },
+    "dependencies": {}
```
Parthiv-M, line 28: Empty entries should be removed

src/Engine/EngineBase.php
```diff
@@ -65,6 +72,7 @@ abstract class EngineBase {
 		'fro' => 'Franceis, François, Romanz (1400-1600)',
 		'ger-hd-m1' => 'Transkribus German handwriting M1',
 		'ger-15' => '15th-16th century German',
+		'german-fraktur-19th-20th-century' => 'German Fraktur 19th-20th century',
```
Parthiv-M, line 75: This is a Transkribus model and needs to be removed from this PR

Parthiv-M (changes requested):
This review focuses on the following main things
- Files related to Transkribus have also been modified, which I think is not ideal
- A new Transkribus model has also been added, which should be removed
- Normalise text has been added here for all engines (we would want that in a separate PR)
- Renaming the newly added API route

Overall, we'd like to keep the non-Kraken changes out of the way before testing kraken once again

src/Controller/OcrController.php
stweil: Having segmentation models is not kraken specific. All OCR processes require a segmentation step, and if that step uses AI, it also requires a model. That's why I did not use a route with "kraken" here. So even if it is currently only used for kraken, I'd suggest to use a generic route.

wrznr:
It is actually a pitty that this PR never got merged. It would have been a great tool for the transcription of texts on wikisource.

stweil:
So let me fix the conflicts first.

samwilson:
I've moved this PR to https://gitlab.wikimedia.org/toolforge-repos/ocr/-/merge_requests/2

stweil:
@samwilson, do I need a new account for gitlab.wikimedia.org? Or is one of my existing accounts also valid there?

samwilson:
You should be able to log in there with your Wikimedia developer account. Let me know if you have any issues though.

I can bring the patch up to date. I feel bad that I dropped the ball on this last year. I'm also thinking that we need to do some Apache proxy config so that kraken requests get forwarded to the right server.

stweil:
I get this message on gitlab.wikimedia.org: "Your account is pending approval from your GitLab administrator and hence blocked. Please contact your GitLab administrator if you think this is an error."

samwilson:
I've added you to trusted-contributors: https://phabricator.wikimedia.org/feed/7635354101851466613/

I think you might need to log in to Phabricator with your developer account, at the moment you don't have a ldap username linked: https://phabricator.wikimedia.org/p/sweil/

PR #90, opened 2023-08-27
Add test cases for Transkribus related methods

kolakachi:
Added a method testFilterValidLangsTranskribusEngine in the EngineBaseTest class.
Added test method for methods related to Transkribus defined in AppExtension.

Bug: T343003:


tests/Engine/EngineBaseTest.php
```diff
@@ -109,6 +125,33 @@ public function testFilterValidLangsTesseractEngine(
...
 	}
 
+	/**
+	 * @param string[] $langs Language codes
+	 * @param string[] $validLangs
+	 * @param string $invalidLangsMode
+	 * @covers EngineBase::filterValidLangs for Transkribus Engine
+	 * @dataProvider provideTranskribusLangs
+	 */
+	public function testFilterValidLangsTranskribusEngine(
```
samwilson: This method looks very similar to the above `testFilterValidLangsTesseractEngine()`, and could perhaps be combined with that and only have one data provider method (which could provide an engine name as well). What do you think?

tests/Engine/EngineBaseTest.php
```diff
@@ -153,4 +213,13 @@ public function testLangNames(): void {
...
 		}
 	}
+
+	/**
+	 * @covers EngineBase::getValidLineIds
+	 */
+	public function testValidLineIds(): void {
+		static::assertNotEmpty( $this->transkribusEngine->getValidLineIds( false, false ), "Missing line IDs" );
+		static::assertNotEmpty( $this->transkribusEngine->getValidLineIds( false, true ), "Missing line ID langs" );
+		static::assertNotEmpty( $this->transkribusEngine->getValidLineIds( true, false ), "Missing line IDs" );
```
samwilson, line 164: No `true, true` case?

samwilson (commented):
Looks good! Just one question about duplication of code.

tests/Engine/EngineBaseTest.php
kolakachi: Both methods have been Combined.

tests/Engine/EngineBaseTest.php
```diff
@@ -89,41 +83,48 @@ public function testFilterValidLangsGoogleEngine(): void {
...
 	 */
-	public function provideTesseractLangs(): Generator {
-		// Format is [ [ langs to test ], [ subset of valid languages ] ]
-		$baseCases = [
-			'all valid' => [ [ 'en', 'fr' ], [ 'en', 'fr' ] ],
-			'one invalid' => [ [ 'foo', 'fr' ], [ 'fr' ] ],
-			// 'equ' is excluded on purpose: T284827
-			'intentionally excluded' => [ [ 'equ' ], [] ],
+	public function provideLangs(): array {
+		return [
```
samwilson, line 111: This is testing only two cases; the old version was testing 3 base cases and each with and without an exception. The new version should be testing those 6 for each engine, for a total of 12 tests. We can fix that in a follow-up though.

PR #91, opened 2023-08-28
Add PHP 8.2 to CI

samwilson:
[T332611](https://phabricator.wikimedia.org/T332611)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,17 +13,21 @@ jobs:
 
     env:
       APP_ENV: test
-    
+
     strategy:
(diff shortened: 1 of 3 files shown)


stweil:
I tried that, too, but CI with PHP 8.2 fails for me, so I had to revert it.

samwilson:
Ah well, it's not a high priority; it'll be a while before we need to use 8.2 in production for this tool.

I noticed that the JS dependencies aren't being fully tested in CI, so have made #101 to fix that (and update a bunch).

Parthiv-M:
Worked alright for me

PR #99, opened 2023-08-28
Fix CSS styles (Wikimedia does not permit font-size with px)

stweil:


stweil:
This fixes a CI failure. See [related article](https://www.mediawiki.org/wiki/Manual:Coding_conventions/CSS#Don't_rely_on_px_unit_based_values).

samwilson:
Looks good! But where's the CI failure? It makes sense that this should be caught by stylelint, but it doesn't seem to have been?

stweil:
The CI failure was reported here:
https://github.com/wikimedia/wikimedia-ocr/actions/runs/5997923435/job/16265249788.

It was triggered for me because my kraken test environment updated `stylelint-config-wikimedia` from ^0.10.3 to ^0.16.1.

samwilson:
Ah right, makes sense! Thanks.

PR #101, opened 2023-08-29
Build assets in CI

samwilson:
Add Node set-up to CI, reading version number from .nvmrc, and
update JS dependencies.


sohomdatta1:
The nvmrc version is pinned on Node12 which is EOL, it might be great to update it to 16 especially since we are going to expose it in CI

samwilson:
Oh, good idea. Done.

PR #103, opened 2023-09-11
Add debugging output for bad cache result

samwilson:
The $result should only ever be an EngineResult, but in some circumstances it's returning an array. This adds some debugging output for this, so we can track down what's going on.

Bug: T345035


Parthiv-M:
Looks good to me!

PR #105, opened 2023-09-13
Add Transkribus job queue

samwilson:
Bug: T346201


samwilson:
This does not yet handle refreshing the session ID correctly.

Parthiv-M:
I believe the `composer.json` is not updated. It is throwing this [error in the CI](https://github.com/wikimedia/wikimedia-ocr/actions/runs/6169780405/job/16744729811?pr=105) as well.

- Got this error initially
```console
Symfony\Component\ErrorHandler\Error\ClassNotFoundError^ {#30
  #message: """
    Attempted to load class "KnpTimeBundle" from namespace "Knp\Bundle\TimeBundle".\n
    Did you forget a "use" statement for another namespace?
    """
  #code: 0
  #file: "./vendor/symfony/framework-bundle/Kernel/MicroKernelTrait.php"
  #line: 132
  trace: {
    ./vendor/symfony/framework-bundle/Kernel/MicroKernelTrait.php:132 { …}
    ./vendor/symfony/http-kernel/Kernel.php:386 { …}
    ./vendor/symfony/http-kernel/Kernel.php:786 { …}
    ./vendor/symfony/http-kernel/Kernel.php:128 { …}
    ./vendor/symfony/framework-bundle/Console/Application.php:168 { …}
    ./vendor/symfony/framework-bundle/Console/Application.php:74 { …}
    ./vendor/symfony/console/Application.php:166 { …}
    ./bin/console:43 {
      › $application = new Application($kernel);
      › $application->run($input);
      › 
    }
  }
}
2023-09-13T19:14:49+05:30 [critical] Uncaught Error: Class "Knp\Bundle\TimeBundle\KnpTimeBundle" not found
```
Running `composer require knplabs/knp-time-bundle:*` fixed it. 

- Furthermore, got this error on the `/transkribus` route
```console
The "format_datetime" filter is part of the IntlExtension, which is not installed/enabled; try running
"composer require twig/intl-extra" in "transkribus.html.twig".
```

Running `composer require twig/intl-extra` fixed it

Parthiv-M (commented):
I think the `/transkribus` route in itself looks good, but can we have the name of the document displayed as well? 

And, maybe, once the job is put into the queue (after the user presses the Transcribe button), we should let the user know what their job ID is?

I feel that implementing either (or both) of the above might make it clearer for the user to identify their OCR jobs.

samwilson:
Good point about the date packages — I'd been playing around with "time ago" formats, but in the end figured it was better to be simpler. Have updated composer.json with the required package, and removed the test one from bundles.php.

> can we have the name of the document displayed as well?

I'm not sure what you mean. There isn't a document name for API jobs, because their images aren't stored. This is the metadata we get about each job:

```php
  'jobId' => string '6124906' (length=7)
  'docId' => int -1
  'pageNr' => int -1
  'pages' => string '1' (length=1)
  'type' => string 'PyLaia Proc Api Decoding' (length=24)
  'state' => string 'FINISHED' (length=8)
  'success' => boolean true
  'description' => string 'Done, duration: 32s 830ms' (length=25)
  'userName' => string 'wikimedia@transkribus.eu' (length=24)
  'userId' => int 146172
  'createTime' => int 1694356085173
  'startTime' => int 1694396118330
  'endTime' => int 1694396151160
  'jobData' => string '#Mon Sep 11 03:35:18 CEST 2023 workDir=/tmp/HTR/PyLaia/trpProd/Decode/pylaiaDecode_6124906 GPU_DEVICE=0 doWordSeg=true keepOriginalLinePolygons=false accountingUnit=PAGES doStoreConfMats=false laConfig={"overwriteExistingLayout"\:true,"doBlockSeg"\:true,"doLineSeg"\:true,"doWordSeg"\:false,"doPolygonToBaseline"\:false,"doBaselineToPolygon"\:true,"doSimplifyPolygons"\:false,"b2pBackend"\:"Legacy","jobImpl"\:"TranskribusLaJob","props"\:{"pars.clustering_legacy_type"\:"default","pars.clustering_method"\:"legacy"}} batchSize=20 accountingModel=creditsOnCompletion clearLines=true doLinePolygonSimplification=true useExistingLinePolygons=false modelId=35909 path=/mnt/processing/K/W/KWZNQETOTHKQFLPMEJVZEEFX'
  'resumable' => boolean false
  'jobImpl' => string 'PyLaiaProcApiJob' (length=16)
  'moduleUrl' => string 'http://read-vm10:8080/PyLaiaModule-trpProd-2.3.0' (length=48)
  'moduleName' => string 'PyLaiaModule' (length=12)
  'moduleVersion' => string '2.3.0' (length=5)
  'started' => string '2023-09-11T03:35:18.33+02:00' (length=28)
  'ended' => string '2023-09-11T03:35:51.16+02:00' (length=28)
  'created' => string '2023-09-10T16:28:05.173+02:00' (length=29)
  'pid' => string '22335' (length=5)
  'batchId' => int 0
  'pageid' => int 0
  'tsid' => int 0
  'parent_jobid' => int 0
  'parent_batchid' => int 0
  'colId' => int -1
  'progress' => int 1
  'totalWork' => int 1
  'nrOfErrors' => int 0
  'modelId' => int 35909
  'nrOfElements' => int 1233
  'priority' => int 0
```

> And, maybe, once the job is put into the queue (after the user presses the Transcribe button), we should let the user know what their job ID is?

That's a great idea. I think it'll need a bit of reworking of the job submission flow, so let's leave it for a subsequent patch. But it'd be great to be able to give people actual info about their place in the queue etc. — and also add that in the on-wiki UI as well (which would mean adding an extra API here).

I've also updated this patch to correctly (I think?) re-login when the session has expired. See what you think.

samwilson:
This needs to handle caching differently, to prevent the session ID being stored in Redis with the other cache data.

Parthiv-M:
> isn't a document name for API jobs, because their images aren't stored

Ah, yes, I forgot about that. Then the current meta data is fine I suppose. We can have the other stuff in a separate patch, like you suggested.

- A minor thing, but I felt the table header _Start delay (in minutes)_ can be right aligned as well, like its rows (comparison below)

   **Present**
Screenshot 2023-09-14 at 13-37-46 WikimediaOCR

   **Suggested**
Screenshot 2023-09-14 at 13-38-07 WikimediaOCR

- Need to update [`CONTRIBUTING.md`](https://github.com/wikimedia/wikimedia-ocr/blob/main/CONTRIBUTING.md#for-transkribus-ocr-engine) with instructions for the new environment variables

The rest of it looks good to me!

samwilson:
> A minor thing, but I felt the table header Start delay (in minutes) can be right aligned as well, like its rows (comparison below)

Done.

> Need to update [CONTRIBUTING.md](https://github.com/wikimedia/wikimedia-ocr/blob/main/CONTRIBUTING.md#for-transkribus-ocr-engine) with instructions for the new environment variables

Very good point. Done.

PR #108, opened 2023-09-21
Fix placeholder text for Transkribus langs

stweil:


stweil:
The language selection text "Leave blank for automatic language detection" is wrong for Transkribus which requires a language model, so don't show that text if that OCR engine is selected.

PR #109, opened 2023-09-21
Avoid runtime error when running without Google keyfile

stweil:


stweil:
This patch allows running wikimedia-ocr without a Google keyfile which makes local and test installations much easier.

samwilson:
Thanks!

PR #111, opened 2023-09-22
Remove Redis config for prod, while we debug

samwilson:


samwilson:
I deployed the new Redis caching, and it is either not connecting or taking a very long time. Switching back to filesystem caching for now, while I investigate more.

PR #112, opened 2023-09-22
Add new OCR parameter to normalize the result text

stweil:


stweil:
Example: Tesseract OCR [with](https://kraken-ocr.wmcloud.org/?image=https%3A%2F%2Fupload.wikimedia.org%2Fwikipedia%2Fcommons%2Fe%2Fe6%2FErsteAusbuergerungsliste.jpg&engine=tesseract&langs%5B%5D=frak2021&normalize=1&segmentation_model=default&psm=3&line_id=null) and [without](https://kraken-ocr.wmcloud.org/?image=https%3A%2F%2Fupload.wikimedia.org%2Fwikipedia%2Fcommons%2Fe%2Fe6%2FErsteAusbuergerungsliste.jpg&engine=tesseract&langs%5B%5D=frak2021&segmentation_model=default&psm=3&line_id=null) normalization.

The normalization works with any OCR engine. The cache always stores the original OCR text. Therefore it is possible to switch to normalized text without a new OCR run.

src/Engine/EngineResult.php
```diff
@@ -35,4 +35,19 @@ public function getText(): string {
 	public function getWarnings(): array {
 		return $this->warnings;
 	}
+
+	/**
+	 * Normalize result by replacing some historic characters
+	 */
+	public function normalize() {
+		$this->text = strtr( $this->text, [
```
stweil, line 43: Some (and more) of these translations could be done with [`Normalizer::normalize( $this->text, Normalizer::FORM_KC )`](https://www.php.net/manual/en/normalizer.normalize.php), but that causes a runtime conflict with the Symfony class which is also called Normalizer.

samwilson: > but that causes a runtime conflict with the Symfony class which is also called Normalizer.

It should work fine as long as you use `\Normalizer` here or `use Normalizer;` at the top, to use the intl extension's one. The Symfony class is a polyfill for that for when the intl extension isn't installed. If you use the former, then don't forget to add that extension to the requirements in `composer.json`.

samwilson (commented):
This looks like a good addition, but note that there's been various discussions over the years about how to normalize OCR output, and not always with huge agreement. Mainly because different Wikisources want to do things differently, and many already have gadgets in place for doing the exact replacements that they want.

For example T278443 fix issue with lines being formatted incorrectly, and T250185 Make Wikisource-OCR handle paragraphs better.

I think there needs to be a way to make this configurable per-project, or perhaps retrieve a config from on-wiki (e.g. a `normalize_config` param could point to a JSON page's URL, where the actual replacement patterns are defined).

PR #114, opened 2023-10-05
Disable form submit during transcription

Parthiv-M:
Disable submit button(s) during
transcription and show visual
indicator for the same

Bug: [T334161](https://phabricator.wikimedia.org/T334161)


sohomdatta1:
LGTM, works in testing.

PR #116, opened 2023-10-13
Move OAuth tokens to cache

kolakachi:
Moved tokens to cache,
Refactored TranskribusClient,
Refactored Test Classes.

Bug: T346306


samwilson (commented):
I'm getting a 500 error with this, when trying to run e.g. http://localhost:8000/?image=http%3A%2F%2Flocalhost%2Fwikimedia%2Fwiki1%2Fimages_wiki1%2Fthumb%2F0%2F0b%2FTest_doc.pdf%2Fpage3-2048px-Test_doc.pdf.jpg&engine=transkribus&langs%5B%5D=ben-print&psm=3&line_id=null

Looking at `$response->getInfo('debug')` in `TranskribusClient::request()` gives:

> WWW-Authenticate: Bearer realm="readcoop", error="invalid_token", error_description="Token is not active"

I've not delved into it much, but it looks like it's might be using a request token as an access token? `getAccessTokenResponse()` is not being used, I think.

Is it working for you?

samwilson:
Oh, and a minor point: the `APP_TRANSKRIBUS_ACCESS_TOKEN` and `APP_TRANSKRIBUS_REFRESH_TOKEN` vars are still in `.env`.

kolakachi:
> I'm getting a 500 error with this, when trying to run e.g. http://localhost:8000/?image=http%3A%2F%2Flocalhost%2Fwikimedia%2Fwiki1%2Fimages_wiki1%2Fthumb%2F0%2F0b%2FTest_doc.pdf%2Fpage3-2048px-Test_doc.pdf.jpg&engine=transkribus&langs%5B%5D=ben-print&psm=3&line_id=null
> 
> Looking at `$response->getInfo('debug')` in `TranskribusClient::request()` gives:
> 
> > WWW-Authenticate: Bearer realm="readcoop", error="invalid_token", error_description="Token is not active"
> 
> I've not delved into it much, but it looks like it's might be using a request token as an access token? `getAccessTokenResponse()` is not being used, I think.
> 
> Is it working for you?

I am not getting the error, unable to find `TranskribusClient::request()`, I think  `request()` in `TranskribusClient` is not a static method.

kolakachi:
> Oh, and a minor point: the `APP_TRANSKRIBUS_ACCESS_TOKEN` and `APP_TRANSKRIBUS_REFRESH_TOKEN` vars are still in `.env`.

Removed.

samwilson:
> I am not getting the error, unable to find TranskribusClient::request(), I think request() in TranskribusClient is not a static method.

That's just a common way of referring to the `request()` method; it doesn't necessarily mean that it's static.

Does it return the correct text result for you?

Oh, and I realise now that I pasted the wrong example link above; that one is using a localhost URL, a better example would be http://localhost:8000/?image=https%3A%2F%2Fupload.wikimedia.org%2Fwikipedia%2Fcommons%2Fthumb%2F1%2F14%2FJohn_Hill_Munday%2527s_notebook_06.jpg%2F1024px-John_Hill_Munday%2527s_notebook_06.jpg&engine=transkribus&langs%5B%5D=en-b2022&psm=3&line_id=null

Does that return a correct text result for you? I'm still getting the "Token is not active" error, even after clearing all caches and refreshing.

CONTRIBUTING.md
```diff
@@ -29,24 +29,11 @@ If you need to make asset changes:
...
 You can [create a free account](https://readcoop.eu/transkribus/?sc=Transkribus) for Transkribus, and get a small number of free credits.
 
-To connect the OCR Tool to your account, run the following CLI command to receive an access token for the Transkribus API:
-
-```console
-$ ./bin/console app:transkribus
-```
-
-It will ask for the *username* and *password* of your Transkribus account, and will output an access token and refresh token.
-All of these should be stored in your `.env.local` file:
```
samwilson, line 39: A comment should be retained here explaining that the username and password need to be added to `.env.local`.

samwilson (approved):
Looks good to me. Small comment about the readme.

PR #119, opened 2023-11-27
Fix crop URL parameter sanitization

samwilson:
Crop width and height can be empty strings, which were being turned into 0.


kolakachi (approved):

Parthiv-M (approved):
Looks alright to me

PR #120, opened 2023-11-30
Retry failed Google requests if the message says to

samwilson:
Check the returned error message for a particular string. This is of course fragile, but there doesn't look to be a better way, and hopefully this is a short term workaround until our thumbnail serving becomes more reliable.

Bug: T332125


samwilson:
This can be tested by just trying with a localhost URL, because then Google will never be able to access it. Whether there's something different going on when Google gets rate limited etc I'm not sure, but all we need is for the returned error message to contain the magic "download it" string. :-P

sohomdatta1:
Tested and works on local setup :)

PR #121, opened 2023-12-08
Add new transkribus model for swe-wikisource

kolakachi:
Bug:T352658

### DIFF
--- a/public/langs.json
+++ b/public/langs.json
@@ -624,6 +624,11 @@
             "htr": 53149
         }
     },
+    "swe-lion-i": {
+        "transkribus": {
+            "htr": 5518
+        }
+    },
     "syr": {
         "tesseract": "syr",
         "google": "syr"
--- a/src/Engine/EngineBase.php
+++ b/src/Engine/EngineBase.php
@@ -93,6 +93,7 @@ abstract class EngineBase {
 		'sk-hand' => 'Handwritten Glagolitic',
 		'sr-latn' => 'Српски (латиница)',
 		'swe-3' => 'Stockholm Notaries 1700 3.0',
+		'swe-lion-i' => 'The Swedish Lion I',
 		'syr' => 'leššānā Suryāyā',
 		'uz-cyrl' => 'oʻzbekcha',
 		'uk-20th-print' => 'Printed Ukrainian 20th century',


tests/Controller/OcrControllerTest.php
```diff
@@ -27,6 +27,7 @@ class OcrControllerTest extends OcrTestCase {
 	 * @covers OcrController::getLangs
 	 * @param string[] $getParams
 	 * @param string[] $expectedLangs
+	 * @return void
```
samwilson: This is correct, but why include it in this PR?

composer.lock
samwilson, line 1: Is this Composer update needed in order to add the new model? It might be better to add it in a separate PR.

samwilson (commented):
Looks fine, but I'm just wondering about mixing the different updates in one PR.

tests/Controller/OcrControllerTest.php
kolakachi: Linting was failing for me.

composer.lock
kolakachi: This was missed. will take it out.

tests/Controller/OcrControllerTest.php
samwilson: Ah right. Does that mean linting is failing on this in current main as well?

kolakachi: yes it's failing on main as well.

`  1x: Method "Psr\Log\LoggerAwareInterface::setLogger()" might add "void" as a native return type declaration in the future. Do the same in implementation "MediaWiki\OAuthClient\Client" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in OcrControllerTest::testGetLang with data set #0 from App\Tests\Controller`

samwilson: Oh interesting, is that showing as an error for you? It's a deprecation warning for me, perhaps you have some other env var set to elevate it?

The warnings I see are:

```
Remaining indirect deprecation notices (4)

  1x: Method "Psr\Log\LoggerAwareInterface::setLogger()" might add "void" as a native return type declaration in the future. Do the same in implementation "MediaWiki\OAuthClient\Client" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in OcrControllerTest::testGetLang with data set #0 from App\Tests\Controller

  1x: The "Symfony\Bridge\Doctrine\Logger\DbalLogger" class implements "Doctrine\DBAL\Logging\SQLLogger" that is deprecated Use {@see \Doctrine\DBAL\Logging\Middleware} or implement {@see \Doctrine\DBAL\Driver\Middleware} instead.
    1x in OcrControllerTest::testGetLang with data set #0 from App\Tests\Controller

  1x: Method "Symfony\Component\Console\Command\Command::execute()" might add "int" as a native return type declaration in the future. Do the same in child class "Doctrine\Bundle\DoctrineBundle\Command\CreateDatabaseDoctrineCommand" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in OcrControllerTest::testGetLang with data set #0 from App\Tests\Controller

  1x: Method "Symfony\Component\Console\Command\Command::execute()" might add "int" as a native return type declaration in the future. Do the same in child class "Doctrine\Bundle\DoctrineBundle\Command\DropDatabaseDoctrineCommand" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in OcrControllerTest::testGetLang with data set #0 from App\Tests\Controller
```

Which I agree we should fix :-) but they should probably be in separate PRs.

samwilson:
It's weird that this is now failing. It looks like something's changed upstream, perhaps it's this issue? https://github.com/symfony/symfony/issues/52844

PR #123, opened 2023-12-29
Add API endpoint to retrieve psm options

kolakachi:
Added an end point for fetching psm options for Tesseract

Bug: [T353824](http://phabricator.wikimedia.org/T353824)

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -312,6 +312,28 @@ class OcrController extends AbstractController {
 	}
 
 	/**
+	 * Get a list of Psms available for use with Tesseract.
+	 *
+	 * @Route("/api/tesseract/available_psms", name="apiPsms", methods={"GET"})
+	 * @OA\Response(response=200, description="List of available psm values and labels, in JSON format.")
+	 * @return JsonResponse
+	 */
+	public function apiAvailablePsms(): JsonResponse {
+		$this->setup();
+		$psms = [];
+		for ( $i = 0; $i <= 13; $i++ ) {
+			array_push( $psms, [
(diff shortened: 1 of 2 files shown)


src/Controller/OcrController.php
```diff
@@ -311,6 +311,28 @@ public function apiAvailableLangsAction(): JsonResponse {
 		] );
 	}
 
+	/**
+	 * Get a list of Psms available for use with Tesseract.
+	 *
+	 * @Route("/api/tesseract/available_psms", name="apiPsms", methods={"GET"})
+	 * @OA\Response(response=200, description="List of available psm values and labels, in JSON format.")
```
Parthiv-M, line 318: Maybe we should add _for tesseract_ here?

Parthiv-M (commented):
Work fine for me. I was wondering whether we should remove the _Automatic page segmentation, but no OSD, or OCR. (not implemented)_ altogether? Ping @samwilson

samwilson:
> I was wondering whether we should remove the _Automatic page segmentation, but no OSD, or OCR. (not implemented)_ altogether?

Probably a good idea, given that it's not meant to be used. I don't think many people have noticed that "(not implemented)"! :-) 

Although if that's removed, I'd suggest moving the PSM loop into a central place, because now we've got it in `templates/_tesseract_options.html.twig` and `OcrController::apiAvailablePsms()`. Actually, centralising it is probably worth doing in this patch (e.g. pass the output of `apiAvailablePsms()` into `OcrController::$params` as is done for `available_line_ids`).

samwilson:
The other bits can be done in a follow-up.

PR #125, opened 2024-01-12
Move Tesseract PSM list to a single place

samwilson:
Rather than iterating through 0..13 in two different places, add a new method to the TesseractEngine to return the canonical list. This means we can exclude number 2 which is not implemented.

Bug: T353824


samwilson:
@kolakachi this is ready for review now I think. Thanks!

kolakachi (approved):
Looks alright to me.

PR #131, opened 2024-08-23
Tesseract: Stop throwing an exception for images with no text

samwilson:
The Tesseract wrapper library we use throws an exception when the command doesn't produce any output (i.e. an output file of size 0). We can't just catch that exception because it's also used for other things, so this adds a check for the contents of the message and returns a warning instead.

Bug: T373161


samwilson:
I think there's probably a nicer way of doing this, so have opened https://github.com/thiagoalessio/tesseract-ocr-for-php/issues/246 — but for now, this is probably fine and it'll stop the log spam.

PR #136, opened 2024-12-13
 adds all missing model codes and updates titles for languages codes …

blackpanther26:
…starting with a and b

### DIFF
--- a/public/models.json
+++ b/public/models.json
@@ -122,67 +122,103 @@
         },
         "ca": {
             "languages": ["ca"],
-            "title": "català"
+            "title": "Català / Catalan / Latn / Supports handwriting"
         },
         "ceb": {
             "languages": ["ceb"],
-            "title": "Cebuano"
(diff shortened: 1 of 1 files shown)


public/models.json
```diff
@@ -1,53 +1,125 @@
...
         },
         "bs": {
             "languages": ["bs"],
-            "title": "bosanski"
+            "title": "bosanski / Bosnian / Latn / Supports handwriting"
         },
+        "btk": {
+            "languages": ["btk"],
+            "title": "Toba–Batak / Batak / Latn / Mapped / Supports handwriting"
+        }
```
reedy: missing a `,`

reedy (changes requested):

public/models.json
blackpanther26: done

samwilson:
Thanks for working on this! Looks good.

PR #141, opened 2025-03-17
Add Open Graph metadata for image and description

osujipius:
Fixes [T349467](https://phabricator.wikimedia.org/T349467)

- Add title, description and image open_graph tags
- Update Description Text

### DIFF
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -3,6 +3,15 @@
     <head>
         <meta charset="UTF-8">
         <title>{% block title %}{{ msg('title') }}{% endblock %}</title>
+        <meta property="og:title" content="{{ msg('title') }}" />
+        <meta property="og:description" content="{{ msg('subtitle') }}"/>
+        <meta property="og:type" content="website" />
+        <meta property="og:url" content="{{ url('home') }}"/>
+        <meta property="og:image" content="{{ absolute_url(asset('build/images/WikimediaOCR-logo.svg')) }}" />
+        <meta property="og:image:width" content="50" />
+        <meta property="og:image:height" content="50" />
+        <meta property="og:image:type" content="image/svg+xml" />
+        <meta property="og:image:alt" content="WikimediaOCR Logo" />
         <link rel="icon" href="{{ asset('favicon.ico') }}" />
         {% block stylesheets %}
             {{ encore_entry_link_tags('app') }}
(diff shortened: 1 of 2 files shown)


templates/base.html.twig
```diff
@@ -3,6 +3,11 @@
     <head>
         <meta charset="UTF-8">
         <title>{% block title %}{{ msg('title') }}{% endblock %}</title>
+        <meta property="og:title" content="{{ msg('title') }}" />
+        <meta property="og:description" content="{{ msg('subtitle') }}"/>
+        <meta property="og:type" content="website" />
+        <meta property="og:url" content="https://ocr.wmcloud.org/"/>
```
samwilson: Should this use `{{ url('home') }}` so that it's correct for the current install location? e.g. https://ocr-test.wmcloud.org/ or local development.

templates/base.html.twig
```diff
@@ -3,6 +3,11 @@
...
+        <meta property="og:image" content="{{ absolute_url(asset('build/images/WikimediaOCR-logo.svg')) }}" />
```
samwilson, line 10: There are some other `image` keys that could be good to add, notably `og:image:alt`, see: https://ogp.me/#structured

samwilson (commented):
Looks good! Thanks for working on this.

templates/base.html.twig
osujipius: @samwilson I have added og:image:alt and og:image:type. Was wondering if I should include og:image:width and og:image:height

templates/base.html.twig
osujipius: I have updated to {{ url('home') }}

templates/base.html.twig
samwilson: Good point. The SVG is displayed at 50 x 50, so those would probably be good dimensions. Do you know if OpenGraph has any info about how to describe SVGs? i.e. because they can scale to whatever, it seems that `width` and `height` are probably most useful to get an aspect ratio.

osujipius: I have added a width and height of 50 x 50.
Unfortunately the width/height properties appear to be the only way to describe SVGs in terms of aspect ratio. 
I can increase from 50x50 to a higher value should we need different results

samwilson: I think 50x50 is fine. We can always change it later after we see how it looks in the wild.

samwilson (approved):
Looks good!

kemsguy7:
@samwilson , please will you be creating newer issues soon?

PR #142, opened 2025-03-17
Add all missing model codes and update all titles

osujipius:
Fixes [T381634](https://phabricator.wikimedia.org/T381634)

- Update and add all missing model codes from A to Z


samwilson:
Thanks!

PR #146, opened 2025-12-10
Ruthen chuhaister

bicolino34:
Adds a new Transkribus model: Ruthenian Chuhaister. Primarily intended for Ukrainian Wikisource, but also supports Belarusian


bicolino34:
@samwilson

samwilson:
Thanks!

PR #147, opened 2026-01-11
Add rotate option to OCR tool

shraddhaa09:
Adds rotation buttons to the OCR interface allowing users to rotate scanned images before running OCR.

Changes:
- Add rotate left/right SVG icon buttons to result page
- Update JavaScript to handle rotation via Cropper.js  
- Extend OcrController to accept and pass rotate parameter
- Update all three engines to apply rotation before OCR processing
- Add rotation to cache key for proper result isolation

Fixes: https://phabricator.wikimedia.org/T413556


.gitreview
```diff
@@ -0,0 +1,6 @@
+[gerrit]
+host=gerrit.wikimedia.org
+port=29418
+project=wikisource/wikimedia-ocr.git
+defaultbranch=main
+defaultremote=gerrit
```
sohomdatta1: This file should not be committed,

package-lock.json
```diff
@@ -4,7 +4,6 @@
     "requires": true,
```
sohomdatta1, line 4: These changes are not required as well and probably stem from a mismatched node version of some kind

sohomdatta1 (changes requested):
Haven't looked too deeply at this but noted a few changes, we shouldn't be adding the .gitreview file and changes to package-lock.json. Those are probably artifacts of some system cofig changes?

shraddhaa09:
Thanks for the review. I’ve removed the unintended `.gitreview` file and reverted unrelated `package-lock.json` changes.

sohomdatta1:
https://github.com/wikimedia/wikimedia-ocr/actions/runs/20956740425/job/60223218062?pr=147 brings up test/lint failures, probably a good idea to fix those as well (to my memory the tests should pass)

assets/images/rotate_left.svg
sohomdatta1, line 1: Please use [Codex icons](https://commons.wikimedia.org/wiki/Codex_icons) instead of these ones

templates/output.html.twig
```diff
@@ -92,13 +92,33 @@
...
+                            </button>
 
                             <input type="submit" value="{{ msg( 'submit-crop' ) }}" class="submit-crop btn btn-info disabled" disabled />
 
                             <input type="hidden" name="crop[x]" value="{% if crop.x is defined %}{{ crop.x }}{% endif %}" />
                             <input type="hidden" name="crop[y]" value="{% if crop.y is defined %}{{ crop.y }}{% endif %}" />
                             <input type="hidden" name="crop[width]" value="{% if crop.width is defined %}{{ crop.width }}{% endif %}" />
                             <input type="hidden" name="crop[height]" value="{% if crop.height is defined %}{{ crop.height }}{% endif %}" />
+
+                            {# NEW: rotate hidden field #}
```
sohomdatta1: Extraneous comment?

shraddhaa09:
Hello @sohomdatta1,
I have addressed the feedback from the previous review and resolved the CI and lint issues. All checks are now passing. Please review the changes at your convenience.

sohomdatta1:
I'm confused here. I tested this locally, but the rotation does not work locally on Google OCR. @shraddhaa09 Did you test the functionality locally before pushing this change?

sohomdatta1:
There are rotation buttons, yes, and I can superficially rotate the image, but the image being sent to Google OCR is not rotated (not only that, when I click the "transcribe area" the rotation is reset while the crops are carried over like expected)

sohomdatta1:
Also, please remove the extraneous "NEW" comments and don't upload changes to the github action.

assets/images/rotate_left.svg
shraddhaa09: @sohomdatta1 
I looked through the Codex icon set for the rotation controls, but I couldn’t find dedicated icons for rotate left or rotate right.
Before updating this further, I wanted to confirm the preferred approach here. Would you recommend:
-reusing existing Codex icons (for example undo/redo)
-switching to text-based buttons for now 
-handling rotation icons through a future Codex addition?
Please let me know which option you’d prefer and I’ll update the implementation accordingly.

Parthiv-M: You could just use the existing codex icons, but I'd feel that getting the PR functional w.r.t the rotation is more of a priority here!

PR #148, opened 2026-01-29
T413556-Add rotate option to OCR tool

Agamya-Samuel:
Add rotate option to OCR tool
- Apply image rotation server-side (Imagine) before cropping and OCR,
  - thread the rotate parameter through the controller and engine layers, and add simple UI controls (output.html.twig, assets/app.js) to rotate via Cropper.js.
- Also include rotate in the cache key so rotated results are cached separately.

Fixes: [T413556](https://phabricator.wikimedia.org/T413556)

I noticed there’s already a PR open (#147) addressing this issue. Since that patch includes some changes that might not be strictly necessary, and it hasn’t had recent updates, I went ahead and explored my own implementation in the meantime.

### DIFF
--- a/templates/output.html.twig
+++ b/templates/output.html.twig
@@ -93,12 +93,22 @@
                                 </button>
                             </div>
 
+                            <div class="btn-group" role="group">
+                                <button type="button" class="btn btn-default rotate-left" title="Rotate left">
+                                    ↺
+                                </button>
+                                <button type="button" class="btn btn-default rotate-right" title="Rotate right">
+                                    ↻
+                                </button>
+                            </div>
+
                             <input type="submit" value="{{ msg( 'submit-crop' ) }}" class="submit-crop btn btn-info disabled" disabled />
 
                             <input type="hidden" name="crop[x]" value="{% if crop.x is defined %}{{ crop.x }}{% endif %}" />
                             <input type="hidden" name="crop[y]" value="{% if crop.y is defined %}{{ crop.y }}{% endif %}" />
                             <input type="hidden" name="crop[width]" value="{% if crop.width is defined %}{{ crop.width }}{% endif %}" />
                             <input type="hidden" name="crop[height]" value="{% if crop.height is defined %}{{ crop.height }}{% endif %}" />
+                            <input type="hidden" name="rotate" id="rotate" value="{% if rotate is defined %}{{ rotate }}{% else %}0{% endif %}" />
                         </div>
                         <div><!-- This div is required for Cropper.js -->
                             <img id="source-image" class="img-responsive" src="{{ app.request.get('image') }}" alt="{{ msg('img-alt-text') }}" />
--- a/assets/app.js
+++ b/assets/app.js
@@ -174,8 +174,10 @@ $(function () {
             y = document.querySelector('[name="crop[y]"]'),
             width  = document.querySelector('[name="crop[width]"]'),
             height = document.querySelector('[name="crop[height]"]'),
+            rotate = document.getElementById('rotate'),
             $modeButtons = $('.drag-mode');
-        new Cropper(img, {
+        let cropperInstance = null;
+        const cropper = new Cropper(img, {
             viewMode: 2,
             dragMode: 'move',
             // Remove double-click drag mode toggling, because we've got buttons for that.
@@ -184,6 +186,10 @@ $(function () {
             autoCrop: width.value > 0 && height.value > 0,
             responsive: true,
             ready () {
+                // Store cropper instance for use in event handlers
+                // `this` is the image element; the cropper instance is on `this.cropper`.
+                cropperInstance = this.cropper;
+                
                 // Make textarea match height of image.
                 $('#text').css({
                     height: this.cropper.getContainerData().height,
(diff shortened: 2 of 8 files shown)


Agamya-Samuel:
# Preview:

https://github.com/user-attachments/assets/43563213-a3f9-4fea-8c58-12b16ebbb44e

Parthiv-M:
Have not taken a proper look at the code but on first glance it seems like it'd be better to use either [OOUI icons](https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=ltr&platform=desktop) or [Codex icons](https://doc.wikimedia.org/codex/latest/icons/all-icons.html).

samwilson:
> Since that patch includes some changes that might not be strictly necessary, and it hasn’t had recent updates, I went ahead and explored my own implementation in the meantime.

That other patch was last updated only a couple of weeks ago, so I'd recommend first asking @shraddhaa09 if they're still working on it, and reviewing their patch to discuss how it can be improved. Always better to help a work-in-progress rather than jumping in with a new patch for someone else's task (although as the task isn't assigned to Shraddha it's technically available, but that's probably just because they aren't familiar with how things are done, I'm guessing).

Agamya-Samuel:
> Have not taken a proper look at the code but on first glance it seems like it'd be better to use either [OOUI icons](https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=ltr&platform=desktop) or [Codex icons](https://doc.wikimedia.org/codex/latest/icons/all-icons.html).

Sure.

templates/output.html.twig
```diff
@@ -93,12 +93,22 @@
                                 </button>
                             </div>
 
+                            <div class="btn-group" role="group">
+                                <button type="button" class="btn btn-default rotate-left" title="Rotate left">
+                                    ↺
```
Agamya-Samuel, line 98: I have used Unicode Character for Rotate Left,

templates/output.html.twig
```diff
@@ -93,12 +93,22 @@
...
+                                </button>
+                                <button type="button" class="btn btn-default rotate-right" title="Rotate right">
+                                    ↻
```
Agamya-Samuel, line 101: I have used Unicode Character for Rotate Right,

Agamya-Samuel:
@Parthiv-M thanks for pointing this out.

Currently I implemented the use of - Unicode Characters (`↺` and `↻`) to denote `Rotate Left` and `Rotate Right`

I gone thru the [Codex icons library](https://doc.wikimedia.org/codex/latest/icons/all-icons.html). Codex currently doesn’t provide a dedicated “rotate” icon.

I also wen thru the [OOUI icons library](https://doc.wikimedia.org/oojs-ui/master/demos/?page=icons&theme=wikimediaui&direction=ltr&platform=desktop).

There is an Icon for `Reload` which can be repurposed, as visually they are the same,

image


The other closest available options in OOUI are the curved arrow icons (undo/redo),

image
image


which can be used to represent rotate-left / rotate-right actions in image editing contexts (if that suits the current needs of [T413556](https://phabricator.wikimedia.org/T413556))

If we find this causes confusion or Codex adds rotation-specific icons later, we can revisit.

Or if there is better way we can handle this ?

Agamya-Samuel:
> > Since that patch includes some changes that might not be strictly necessary, and it hasn’t had recent updates, I went ahead and explored my own implementation in the meantime.
> 
> That other patch was last updated only a couple of weeks ago, so I'd recommend first asking @shraddhaa09 if they're still working on it, and reviewing their patch to discuss how it can be improved. Always better to help a work-in-progress rather than jumping in with a new patch for someone else's task (although as the task isn't assigned to Shraddha it's technically available, but that's probably just because they aren't familiar with how things are done, I'm guessing).

Thanks @samwilson for the suggestion. I agree that building on existing work is generally the better path when someone is actively progressing on a patch.

That said, when I looked at the open patch, it included additional changes beyond what was needed for this specific issue, and there wasn’t any clear indication yet whether it was still being actively driven to completion (last commit was on 16 Jan).

Rather than letting the task stall, I explored a minimal, focused implementation in parallel to keep momentum and provide an alternative that can be evaluated.

I’m **more than happy** to coordinate with @shraddhaa09 and review their approach as well. If their patch is still in progress and closer to the intended direction, we can consolidate efforts or adjust accordingly. My goal here isn’t to duplicate work, but to ensure the rotation support lands cleanly and without unnecessary delay.

shraddhaa09:
Hi ,
Just to clarify from my side: I am actively working on PR #147 and addressing the rotation logic, crop reset issue, and UI feedback that’s been raised. I’ll be pushing updates shortly.

Thanks for checking in.

samwilson:
@shraddhaa09 @Agamya-Samuel: Okay great, sounds good! So I think if you two can settle on one of these patches and then work together to review it, that'd be best.

okerekechinweotito:
@shraddhaa09 @Agamya-Samuel 
Please checking to see if you are still working on this issue

Agamya-Samuel:
> @shraddhaa09 @Agamya-Samuel Please checking to see if you are still working on this issue

@okerekechinweotito  I am still in the process of completing the patch

PR #149, opened 2026-01-29
Implement uselang query parameter to set UI language

Agamya-Samuel:
- Added parsing of `?uselang=` in `src/Controller/OcrController.php`.
- Sanitize value to allow only letters, digits, dash and underscore.
- Set Intuition language if possible; try `setLang`.
- Purpose: let URLs override UI language while remaining backward-compatible; no persistence added.

Fixes: [T365940](https://phabricator.wikimedia.org/T365940)

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -90,6 +90,17 @@ class OcrController extends AbstractController {
 	 * @suppress PhanSuspiciousValueComparison
 	 */
 	private function setup(): void {
+		// Parse `?uselang=` and set Intuition language (sanitized). Try multiple setters for compatibility.
+		$uselang = $this->request->query->get( 'uselang' );
+		if ( $uselang ) {
+			// Sanitize to alphanumerics, dash and underscore only.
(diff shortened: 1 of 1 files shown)


Agamya-Samuel:
# Preview:

https://github.com/user-attachments/assets/e465976f-eae2-44a5-8cc6-ba5044475d7e

Agamya-Samuel:
earlier I implemented multiple setters, to work with diff version of the [krinkle/intuition](https://packagist.org/packages/krinkle/intuition): 

image

---
But now I have made changes according to the latest version of krinkle/intuition: v2.3.6

image

which defines `setLang()`

samwilson:
Thanks for working on this! I'm not sure that this is the correct fix. The ToolforgeBundle is already [setting the language](https://github.com/wikimedia/ToolforgeBundle/blob/1663a9ccafc007fe7fd6f45d014e3bec4dc907dd/Service/Intuition.php#L56) for Intuition. The issue as I understand it is that we need to upgrade the version of ToolforgeBundle that's used here.

samwilson:
Superseded by #150.

PR #153, opened 2026-02-11
Add new Ukrainian model

bicolino34:
Added new Ukrainian model (Ukrainian generic handwritten and typed)


bicolino34:
@samwilson

samwilson:
Thanks!

PR #157, opened 2026-02-28
T383002 | Format language arrays in models.json to multi-line style

vm-pranavan:


samwilson:
Thanks for this! I assume this is format is wanted in order to make it easier to update the models.json file programmatically. If so, it'd be good to explain that in the PR description (see also the [commit message guidelines](https://www.mediawiki.org/wiki/Gerrit/Commit_message_guidelines), including how to reference the task number).

Also, there are some unrelated changes that should be done in a separate PR.

vm-pranavan:
Hi @samwilson 

Thanks for the review!

The unrelated formatting changes have been removed — the PR now only contains the two intentional changes:

1. Adding mnp (Northern Min) to models.json
2. Switching textDetection → documentTextDetection for better diacritic recognition

The commit message has also been updated to reference the task: Bug: T383002

samwilson:
Will be back at my desk tomorrow so will review then. I seem to remember switching to document-text-detection was talked about (and done!) before: https://phabricator.wikimedia.org/T246944#5978037 (not sure why we switched back to text-detection).

vm-pranavan:
Hi @samwilson 

Thanks for the context! It looks like documentTextDetection was reverted unintentionally at some point — this PR restores the original intended behaviour.

samwilson:
Thanks!

PR #166, opened 2026-03-04
T411447: Fix the CORS error when fetching https://ocr.wmcloud.org/models.json

okerekechinweotito:
Related Phabricator Task - [T411447](https://phabricator.wikimedia.org/T411447)
Related Gerrit Patch - https://gerrit.wikimedia.org/r/c/mediawiki/extensions/Wikisource/+/1247993

This PR fixes the CORS error when fetching  [https://ocr.wmcloud.org/models.json](https://ocr.wmcloud.org/models.json) from inside the Wikisource extension.

I attempted to fetch https://ocr.wmcloud.org/models.json directly from the frontend. However, because this file is served as a static asset by nginx (not by Symfony/PHP), it did not include the required Access-Control-Allow-Origin header, resulting in a CORS error in browsers.

To resolve this, the PR introduces a new Symfony route `/api/models` that reads and returns the contents of models.json with the correct CORS headers. This ensures the model data is accessible to the frontend, matches the expected data structure, and is served with the necessary CORS headers, while leaving the static file untouched.

### DIFF
--- a/src/Controller/OcrController.php
+++ b/src/Controller/OcrController.php
@@ -295,6 +295,28 @@ class OcrController extends AbstractController {
 	}
 
 	/**
+	 * Serve the models config with CORS headers so it can be fetched directly
+	 * from the Wikisource frontend without requiring a MediaWiki proxy.
+	 * Returns a simplified map of engine->{code: title} matching the shape
+	 * previously provided by the available_langs API.
+	 *
+	 * @Route("/api/models", name="apiModels", methods={"GET"})
+	 * @return JsonResponse
+	 */
+	public function apiModelsAction(): JsonResponse {
+		$path = $this->getParameter( 'kernel.project_dir' ) . '/public/models.json';
+		$raw = json_decode( file_get_contents( $path ), true );
+		$out = [];
+		foreach ( $raw as $engine => $models ) {
+			$out[ $engine ] = [];
+			foreach ( $models as $code => $info ) {
+				$out[ $engine ][ $code ] = $info[ 'title' ];
+			}
+		}
+		return $this->getApiResponse( $out );
+	}
+
+	/**
 	 * Get a list of models available for use with a specific OCR engine.
 	 *
 	 * @Route("/api/available_langs", name="apiLangs", methods={"GET"})
--- a/config/packages/nelmio_api_doc.yaml
+++ b/config/packages/nelmio_api_doc.yaml
@@ -7,6 +7,4 @@ nelmio_api_doc:
     areas:
         path_patterns:
             - ^/api$
-            - ^/api/available_langs$
-            - ^/api/tesseract/available_psms$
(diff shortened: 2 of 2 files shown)


samwilson:
Sorry, it might not have been obvious in the task, but it's the Wikisource extension (for MediaWiki) that is making the request through MediaWiki. No change is needed to the tool as far as I know.

okerekechinweotito:
@samwilson 
Sorry if the PR title was a bit confusing. This PR is to fix the CORS error I get when I call  `https://ocr.wmcloud.org/models.json` from inside the wikisource extension. 
I have already made a patch to wikisource (gerrit link attached to this PR description) that enables direct calls to  `/models.json`

samwilson:
Ah great, yes I see now! Makes sense.

I'll try to review these tomorrow. Thanks for working on it all!

src/Controller/OcrController.php
```diff
@@ -294,6 +294,28 @@ public function apiAction(): JsonResponse {
...
+	 * @return JsonResponse
+	 */
+	public function modelsJsonAction(): JsonResponse {
+		$path = $this->getParameter( 'kernel.project_dir' ) . '/public/models.json';
+		$raw = json_decode( file_get_contents( $path ), true );
+		$out = [];
+		foreach ( $raw as $engine => $models ) {
+			$out[ $engine ] = [];
+			foreach ( $models as $code => $info ) {
+				$out[ $engine ][ $code ] = $info[ 'title' ] ?? '';
```
samwilson: The `title` key is required, so I don't think there's any need to fall back to empty string.

src/Controller/OcrController.php
```diff
@@ -294,6 +294,28 @@ public function apiAction(): JsonResponse {
 		return $this->getApiResponse( $responseParams );
 	}
 
+	/**
+	 * Serve the models config with CORS headers so it can be fetched directly
+	 * from the Wikisource frontend without requiring a MediaWiki proxy.
+	 * Returns a simplified map of engine->{code: title} matching the shape
+	 * previously provided by the available_langs API.
+	 *
+	 * @Route("/api/models", name="modelsJson", methods={"GET"})
```
samwilson: Suggest calling it `apiModels` in line with the other API methods (see `./bin/console debug:router --env prod` for a list).

Also, it looks like the regex in `config/packages/nelmio_api_doc.yaml` needs to be updated so that this new one is listed in the Swagger docs. I'm not sure why it has all the endpoints listed separately there, if it was just `/api/.` it'd probably catch them all (but not the legacy `api.php` one).

samwilson (commented):
Looks great! Just a couple of little comments.

src/Controller/OcrController.php
okerekechinweotito: updated the name and made the switch to ^/api/

PR #167, opened 2026-03-17
Fix: handle array input for engine parameter in OcrController

monicadola42:
Fixes error when 'engine' parameter is passed as an array.

Previously, passing engine[]=tesseract caused a TypeError because EngineFactory::get() expects a string.

This change ensures that if the 'engine' parameter is an array, the first value is extracted before passing it to the factory.

Tested locally using PHP built-in server.

PR #168, opened 2026-03-17
Fix: handle array input for engine parameter to prevent TypeError

monicadola42:
## Description

This PR fixes an issue where passing the `engine` parameter as an array (e.g., `engine[]=tesseract`) caused a TypeError in the application.

## Changes made

- Added handling for array input in the `engine` parameter
- Ensured the first value is safely extracted when an array is provided
- Prevented application crash by validating input type

## Testing

Tested the following cases locally:

- `engine=tesseract` → works correctly
- `engine[]=tesseract` → handled safely without crashing

## Notes

This ensures the application behaves correctly even when the `engine` parameter is passed incorrectly as an array.


monicadola42:
Closing this PR as suggested. Continuing updates in #167.

PR #176, opened 2026-03-31
T421660: eslint enforce semi-style

okerekechinweotito:

- remove `semi-style` from .eslintrc.json
- fix `semi-style` errors
- `npm run test` passes
- verify that there are no breaks in functionality


webpack.config.js
```diff
@@ -72,6 +72,5 @@ Encore
 
 // uncomment if you're having problems with a jQuery plugin
 // .autoProvidejQuery()
```
samwilson: These comments are not really needed, and can be removed as well.

okerekechinweotito: done

PR #177, opened 2026-03-31
T421660: eslint enforce yml/no-empty-document

okerekechinweotito:

the yaml documents have commented out content in them. I am not sure deleting the files will be an accepted option so I have added the `~` character to make eslint see them as `non-empty`

- remove `yml/no-empty-document` from .eslintrc.json
- fix `yml/no-empty-document` errors 
- `npm run test` passes

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,7 +12,6 @@
 		"sourceType": "module"
 	},
 	"rules": {
-		"no-jquery/no-global-selector": "warn",
-		"yml/no-empty-document": "warn"
+		"no-jquery/no-global-selector": "warn"
 	}
 }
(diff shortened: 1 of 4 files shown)


samwilson:
Are these three files needed at all? If we don't have any config for them, it might be best to remove them. What do you think?

okerekechinweotito:
> Are these three files needed at all? If we don't have any config for them, it might be best to remove them. What do you think?

I agree i think the files can be safely deleted. for any future config their use is well documented, i found some docs references to them if need be

webpack_encore.yaml
- [webpack_encore.yaml config showing strict_mode and cache](https://symfony.com/bundles/WebpackEncoreBundle/current/index.html#configuration)

deprecations.yaml
- [reference of deprecations.yaml file in symfony recipes (github)](https://github.com/symfony/recipes/blob/main/symfony/monolog-bundle/3.1/config/packages/prod/deprecations.yaml)
- [Symfony Logging (Monolog)](https://symfony.com/doc/current/logging.html#monolog)
- [log messages to different files](https://symfony.com/doc/current/logging/channels_handlers.html)

samwilson:
Thanks!

PR #179, opened 2026-04-06
T421660: eslint enforce no-jquery/no-global-selector

okerekechinweotito:

- remove `no-jquery/no-global-selector` from .eslintrc.json
- fix `no-jquery/no-global-selector` errors 
- `npm run test` passes

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -3,8 +3,10 @@ import './styles/app.css';
 import 'select2';
 
 const $ = require( 'jquery' );
+// eslint-disable-next-line no-jquery/no-global-selector
 const $select2 = $( '#lang' );
 let selectedLanguages = [];
+// eslint-disable-next-line no-jquery/no-global-selector
 const $lineDetectionSelect = $( '#line-id' );
 let lineModels = null;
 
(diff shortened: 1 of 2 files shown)


assets/app.js
```diff
@@ -3,9 +3,10 @@ import './styles/app.css';
 import 'select2';
 
 const $ = require( 'jquery' );
-const $select2 = $( '#lang' );
+const $document = $( document );
```
samwilson: Although this does get rid of the `no-jquery/no-global-selector` errors, it doesn't really get to the point of that rule. It's just hiding the fact that there are global selectors being run.

The idea is that we shouldn't be querying the entire DOM for each element that's needed, but rather just fetching either by ID (which is more efficient), by e.g. fetching some parent element and then doing `find()` within just that element, or maybe by getting rid of jQuery altogether and using `document.querySelector()` (which is already done for some).

samwilson (changes requested):

assets/app.js
okerekechinweotito: okay, i switched to `getElementById` , `querySelector` and `querySelectorAll`
i am on standby for any followup changes needed

samwilson: I'm not sure that just repeating the same queries but with querySelector quite gets to the root of why this rule exists. For example, rather than changing `$( '[name=image]' )` to `$( document.querySelector( '[name=image]' ) )` it's probably better to first find the form with e.g. `$( 'form.ocr-form' )` and then search within that for the `image` field. However, in this case it's probably easier to put a one-line rule exclusion in!

Wrapping `querySelector()` or `getElementById()` in jQuery feels redundant, as that's just what jQuery will do internally for these simple selectors anyway. There are a few, such as removing the `nojs` class, where you probably don't need jQuery at all.

samwilson (approved):
Looks good! Thanks.

